PR #5, opened 2026-05-04
M0: skeleton

spxrogers:
## Summary

Implements the M0 skeleton milestone for agentsync per the plan at `docs/superpowers/plans/2026-05-04-agentsync-m0-skeleton.md`.

**What shipped (Batches A–D, Tasks 1–17):**
- `internal/paths` — AGENTSYNC_HOME resolution with AGENTSYNC_TARGET_ROOT override for tests
- `internal/iox` — atomic file write (tmp+fsync+rename) and cross-process flock
- `internal/log` — slog JSON logger with `--verbose` toggle
- `internal/source` — canonical schema (TOML structs) + loader that walks `~/.agentsync/`
- `internal/state` — targets.json read/save with atomic write and schema v1
- `internal/adapter` — Adapter interface + in-memory Registry
- `internal/adapter/noop` — NoopAdapter for testing the apply pipeline pre-M1
- `internal/render` — Plan + Apply pipeline orchestrator (RenderPlan)
- `internal/cli` — Cobra root with `--verbose`; five subcommands:
  - `agentsync init` — scaffold `~/.agentsync/` with empty subdirs + `agentsync.toml`
  - `agentsync agent add/list/remove` — TOML round-trip mutation of `[agents]` section
  - `agentsync doctor` — environment + PATH-based agent detection
  - `agentsync verify` — schema-lint every TOML file + validate agent names
  - `agentsync apply --dry-run` — source → render.Plan → printed summary (NoopAdapters)
- `cmd/agentsync/main.go` — wired to `cli.Execute()`

## Test plan

- [x] `go build ./...` passes
- [x] `go test -race ./...` — all 9 packages pass (no failures, no races)
- [x] `go vet ./...` — clean
- [x] `agentsync --version` prints `agentsync dev (commit none, built unknown)`
- [x] `AGENTSYNC_TARGET_ROOT=/tmp/m0-demo agentsync init` → `agentsync home initialized at /tmp/m0-demo/.agentsync`
- [x] `agentsync agent add claude` → `added agent: claude`
- [x] `agentsync agent add opencode` → `added agent: opencode`
- [x] `agentsync agent list` → `claude     enabled=true scope=user` + `opencode   enabled=true scope=user`
- [x] `agentsync verify` → `ok: schema valid; all referenced agents are recognized adapters`
- [x] `agentsync apply --dry-run` → `Plan: 0 ops total across 2 agent(s)` with per-agent breakdown
- [x] `agentsync apply` (no --dry-run) → error: `M0 only supports --dry-run; real adapters arrive in M1 (claude) and M2 (opencode)`
- [x] `agentsync doctor` → prints AGENTSYNC_HOME, Go version, OS, per-adapter PATH detection
- [x] Integration test `TestIntegration_M0Lifecycle` covers full lifecycle in tmpdir

## Notable deviations from plan

- Config file is `agentsync.toml` (not `opensync.toml`) per design spec canonical
- `render.Plan` struct renamed to `render.RenderPlan` to avoid Go name collision with `Plan()` function (from Batch C)
- `writeAgents` uses inline TOML format (`claude = { enabled = true, scope = "user" }`) instead of go-toml encoder sub-tables — necessary because the encoder emits `[agents.claude]` which causes duplicate-table parse errors on subsequent writes


cmd/agentsync/main.go
```diff
@@ -0,0 +1,25 @@
...
+	"os"
+
+	"github.com/spxrogers/agentsync/internal/cli"
+)
+
+var (
+	version = "dev"
+	commit  = "none"
+	date    = "unknown"
+)
```
spxrogers, line 14: when / what process will set these `var` values?

spxrogers: [goreleaser](https://goreleaser.com/) injects them at release-build time via Go's `-ldflags -X` mechanism. Specifically, [`.goreleaser.yaml:10-14`](https://github.com/spxrogers/agentsync/blob/claude/agentsync-m0/.goreleaser.yaml#L10-L14):

```yaml
ldflags:
  - -s -w
  - -X main.version={{.Version}}
  - -X main.commit={{.ShortCommit}}
  - -X main.date={{.Date}}
```

When goreleaser builds a release, it overwrites the package-level `version` / `commit` / `date` vars in `main` with the tagged version, the short commit SHA, and the build timestamp.

For non-release builds (`just build`, `go build`, anything CI runs that isn't `goreleaser release`), the defaults stand — so `agentsync --version` prints `agentsync dev (commit none, built unknown)`. That's intentional: local dev builds shouldn't lie about being a real version.

The flow:
1. ldflags overwrite `main.version` / `main.commit` / `main.date` (or leave defaults).
2. `main.go:17-19` forwards them into the `cli` package's exported `Version` / `Commit` / `Date` vars.
3. `internal/cli/root.go:27,29` consumes them via cobra's `--version` template.

Worth noting that `internal/cli/root.go:12-14` declares its own `"dev" / "none" / "unknown"` defaults too — so even if `main.go` didn't forward, `--version` would still produce a valid (always-"dev") string. Belt and suspenders.

CI's `goreleaser release --snapshot --skip publish --clean` exercises this path on every push, so the ldflag injection is continuously verified. The first real release (when [LAUNCH_CHECKLIST.md §3](https://github.com/spxrogers/agentsync/blob/claude/launch-checklist/LAUNCH_CHECKLIST.md) gets worked through and a `v1.0.0` tag is pushed) will produce the real values for the first time.

PR #20, opened 2026-05-16
Pre-launch hardening: 19 audit findings (secrets, symlinks, state portability, supply chain)

spxrogers:
## Summary

Applied a superpowers-style audit (read-only, paranoid, root-cause oriented) across the whole repo, then landed fixes for every finding before first real-machine use. Eight focused fix commits, ~5.3k lines changed across 77 files, full `go test -race ./...` green.

The audit assumed the user was about to point this at their real `~/.claude.json`, real Anthropic/GitHub tokens, and a real chezmoi-synced dotfiles tree — so the bar was "what bites on first run" rather than "what fails in a controlled test."

## Showstoppers (would-have-bit-on-first-run)

| Finding | Fix | Commit |
| --- | --- | --- |
| AtomicWrite silently replaced symlinked destinations with regular files (the README's own chezmoi recommendation triggers this) | Refuse by default; `AGENTSYNC_ALLOW_SYMLINK_DEST=1` resolves through the link in place so the symlink survives | `517a6c1` |
| State keys embedded absolute dest paths → chezmoi sync from `/Users/alice/` to `/home/alice/` reclassified every native file as ForeignCollision | `paths.HomeRelative` normalization at every key construction/lookup site | `9ac6e80` |
| `apply --dry-run` only printed op counts — the README's "preview before writing" hid the most destructive thing apply does | Dry-run invokes the writer's collision detector in no-write mode and lists every destination path + planned backup | `e3e824c` |
| `agentsync diff` printed resolved secret values to stdout (dest was written by prior apply with secrets substituted as cleartext) | `secrets.CollectResolved` + `MaskResolved` redact every resolved value back to its `${secret:…}` placeholder on both sides | `e3e824c` |
| `secrets set github.token=ghp_REAL` exposed the value to `ps(1)`, `/proc/<pid>/cmdline`, `.bash_history`, auditd | `--stdin` flag + `term.ReadPassword` interactive prompt; legacy `=value` warns; malformed args no longer echo their value | `5784c78` |
| `plugin disable` was decorative — the loader's `PluginSpec` had no `Disabled` field, so the bit was silently dropped on every Load | Added `Disabled` to source schema; `projectPlugins` honors it | `9ac6e80` |
| Mid-apply failure (write #5 of 10 errors) left files 1-4 on disk but never called `state.Save` → next apply backed them up as foreign | `saveBestEffortState` records hashes for ops whose dest exists, then propagates the original error | `ede7d60` |
| `secrets.Encrypt` used `O_TRUNC` then streamed — Ctrl-C between truncate and close nuked every key with no backup | Buffer encrypted bytes in memory, commit via `iox.AtomicWrite` | `5784c78` |

## Other logic gaps

| Finding | Fix | Commit |
| --- | --- | --- |
| Reconcile `[w]rite-back` for hooks/lsp/multi-source items returned nil with a "write-back: <label>" success message — silent data loss | Explicit errors for unsupported pointer shapes, missing SourceID, and `(multiple)` SourceIDs | `a848d2c` |
| Reconcile filtered out ForeignCollision items — exactly the bootstrap case where user most needs the warning | Include ForeignCollision in `requiresAction` | `a848d2c` |
| Bulk-action W/O/S locked in for all remaining items with no preview, no cancel — one stray shift-W could wipe edits across the queue | "apply 'w' to all N remaining items? [y/N]" confirmation, default no | `a848d2c` |
| `import` had no global lock — concurrent `apply` ↔ `import` race lost state writes | `withGlobalLock` wrapper | `af344fe` |
| `import` silently left non-imported dest pointers as ForeignCollision time bombs | Warns at import time, lists each unowned pointer with remediation pointer | `ede7d60` |
| Managed file replaced with symlink invisible to `status` (os.ReadFile follows the link) | `hashFile` Lstats first; symlink yields sentinel hash that never matches → classifies as Drift | `af344fe` |
| No size limit on npm tarball decompression — gzip-bomb fills disk | Per-tarball decompressed-bytes cap (default 512 MiB, override `AGENTSYNC_MAX_TARBALL_MB`); metadata response capped at 16 MiB; loud `cappedReader` instead of silent io.LimitReader EOF | `af344fe` |
| `ManifestSHA` recorded at install but never verified at apply → cache tampering silent | `verifyPluginManifestSHA` on every projection; `AGENTSYNC_ALLOW_PLUGIN_DRIFT=1` opt-in | `ede7d60` |
| `marketplace add gh:typo/plugin` silently `cp -r ./gh:typo/plugin` creating an empty marketplace | Unknown URL forms error with the list of supported shapes; bare local dirs still work | `af344fe` |
| AtomicWrite tmp file created at caller mode (0644 by default) — secrets briefly world-readable in destination dir during rename window | Tmp always 0600; chmod to target mode after rename | `517a6c1` |
| `verify` had its own inline `runtime.GOOS != "windows"` permission check that ignored `AGENTSYNC_AGE_SKIP_PERM_CHECK` — apply and verify disagreed on "secure" | Both use `secrets.CheckIdentityPermissions` | `af344fe` |

## Test coverage added

- `TestDiff_DoesNotLeakResolvedSecrets` — sentinel token end-to-end
- `TestAtomicWrite_RefusesSymlinkDest` / `_AllowsSymlinkDestWithEnv` / `_TmpFileIsNotWorldReadable`
- `TestApply_DryRunPreviewsForeignCollisions` / `_DryRunListsDestinations`
- `TestSecretsSet_NoEqualsNonTTYErrors` / `_StdinPath` / `_LegacyArgWarns`
- `TestPlugin_DisableSuppressesProjectionAtApply`
- `TestLoad_PluginManifestSHAMismatchRefuses` / `_OverrideEnv`
- `TestNPMFetcher_GzipBombIsCapped`
- `TestReconcile_BulkActionRequiresConfirmation` / `_WriteBackUnsupportedReturnsError`
- `TestState_PortableAcrossHomes` + `TestHomeRelative` unit cases
- Plus `secrets.MaskResolved` / `CollectResolved` unit suite (sentinel, longest-match-first, nil-map no-op, unresolved-skip)

## Deliberately not addressed (with reasoning)

- **Mode-comparison drift** — chmod 600 on `.claude.json` by security-conscious users is legitimate; flagging it as drift generates noise. The dangerous case (symlink swap) IS now detected.
- **Concurrent-apply real-fork test** — `gofrs/flock` uses fd-level `flock(2)` on Linux; the existing same-process test exercises the same primitive cross-process apply uses.
- **Crash-during-apply chaos test** — fault-injection harness deferred to v1.x; `saveBestEffortState` is the runtime mitigation.
- **`AGENTSYNC_ALLOW_OFFLINE_VERIFY` typo detection** — offline mode cannot catch missing secret keys without a backend; the regex inside `SubstituteRefs` already enforces reference shape.

## Test plan

- [x] `go test -race ./...` green (full suite, all packages)
- [ ] `just test-release` (hermetic container release gate) — needs maintainer to run; my environment lacks podman/docker
- [ ] Manual smoke on a real machine: `init`, `agent add claude`, `apply --dry-run` against a populated `~/.claude.json`, observe the new foreign-collision preview
- [ ] `secrets set github.token --stdin` from a clean shell, then `apply`, then `diff` and confirm the token doesn't appear in output

## Files

77 files changed, 5,304 insertions, 350 deletions. Eight fix commits + one gofmt cleanup.

https://claude.ai/code/session_01Syn3eQdUTc6wGEh8RcSwJp

---
_Generated by [Claude Code](https://claude.ai/code/session_01Syn3eQdUTc6wGEh8RcSwJp)_

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -58,6 +58,11 @@ linters:
         linters: [forbidigo]
       - path: '(^|/)internal/state/store\.go$'
         linters: [forbidigo]
+      # secrets/age.go writes secrets.age under ~/.agentsync/ via the
+      # atomic path so a Ctrl-C during re-encrypt can't truncate the file.
+      # Same canonical-source category, not a native destination.
+      - path: '(^|/)internal/secrets/age\.go$'
+        linters: [forbidigo]
       # CLI files writing to ~/.agentsync/* (canonical source).
       - path: '(^|/)internal/cli/(plugin|marketplace|agent|reconcile|update)\.go$'
         linters: [forbidigo]
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ agentsync is single-machine. To sync `~/.agentsync/` across machines, use chezmo
 
 If you lose your age private key, you lose access to all encrypted secrets. Recommended: store the key in a 1Password Secure Note or your machine-setup repo. agentsync does not back up the key for you.
 
+`agentsync secrets set` accepts the value three ways:
+
+    agentsync secrets set github.token --stdin    # value from stdin (best for scripts / 1Password CLI)
+    agentsync secrets set github.token            # prompt with echo off
+    agentsync secrets set github.token=ghp_…      # back-compat; warns — argv is visible to ps(1) and shell history
+
+`agentsync diff` redacts every resolved `${secret:…}` value before printing, so a piped diff doesn't leak credentials to logs.
+
 ## Known limits in v1.x
 
 - **OpenCode hooks**: OpenCode hooks are JS/TS plugins, not declarative shell commands. agentsync v1 does NOT auto-translate Claude hooks to OpenCode. Hand-author a small JS/TS plugin if you need a hook on OpenCode.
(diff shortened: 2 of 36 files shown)


spxrogers:
Superseded by #22 — GitHub auto-closed this PR (and blocked reopen) when #19's base branch was deleted during the stacked squash-merge. The identical work, rebased onto `main`, is in #22.

PR #73, opened 2026-06-06
fix: skip unsupported Claude LSP settings

### LINKED ISSUE #66: claude adapter: LSP servers are written to settings.json#/lspServers, but Claude reads LSP only from plugin manifests
## Summary

The Claude adapter renders **and** ingests LSP servers at `settings.json#/lspServers` (`internal/adapter/claude/lsp.go`, `internal/adapter/claude/ingest.go`), at both user and project scope. But per the upstream Claude Code docs, **`lspServers` is not a `settings.json` key at all** — Claude Code reads LSP servers **only from plugin manifests** (`.lsp.json` at a plugin root, or inline in `plugin.json` under `lspServers`).

So an LSP server agentsync writes into `settings.json` is **silently ignored by Claude Code** (unrecognized key), and `import`/`reconcile` will never find one there because Claude never writes one there. Unlike the project-MCP bug (#56), this is not a user-vs-project misplacement — it's a wrong-file-entirely placement at *every* scope: LSP doesn't live in `settings.json`.

This contradicts the repo's own capability matrix, which claims `LSP server | ✓ native` for Claude (`docs/capability-matrix.md:72`) and "All seven components, including LSP" (`:23`), plus the `CapLSP` bitmask (`internal/adapter/claude/claude.go`). Per CLAUDE.md's "doc claims are not self-certifying" and "models must stay faithful to their on-disk artifacts" rules, the native-LSP claim is unbacked.

> ⚠️ Per `CLAUDE.md` ("Cross-reference upstream harness docs before guessing"), confirm the LSP config location against https://code.claude.com/docs/ before implementing any remediation. The evidence below was gathered during the #56 fan-out audit, but re-verify.

## Evidence (verified against upstream during the #56 audit)

- The [settings reference](https://code.claude.com/docs/en/settings) enumerates ~93 top-level `settings.json` keys; `hooks` is present, **`lspServers` is not** among them.
- The [plugins reference](https://code.claude.com/docs/en/plugins-reference) documents `lspServers` **only** as a plugin component:
  - *"Plugin components include skills, agents, hooks, MCP servers, **LSP servers**, and monitors."*
  - **Location**: `.lsp.json` in plugin root, or inline in `plugin.json` (`"lspServers": "./.lsp.json"`).
  - The `plugin.json` schema row: `lspServers | string|array|object | LSP configs … | "./.lsp.json"`.

## Current behavior — exact locations

- **Render** — `internal/adapter/claude/lsp.go` emits a `settings.json` FileOp with `/lspServers/<id>` (merge-json-keys), mirroring MCP/hooks.
- **Ingest** — `internal/adapter/claude/ingest.go` reads `settings.json` `/lspServers/<id>` back into `source.LSPServers`.
- Round-trip is internally consistent (render and ingest agree on the wrong file), so no existing test fails — every LSP test asserts the `settings.json` shape, not that Claude Code actually loads it. (cf. the same "impoverished oracle" failure mode called out in CLAUDE.md.)

## Remediation options (decision needed — do not just patch a path)

There is **no `settings.json` slot to move LSP to**; the only native LSP surface is a plugin. So this is a design decision, not a path move:

1. **Downgrade Claude LSP to a documented Skip (`◐`/`✗`).** Stop emitting `settings.json#/lspServers`; drop `CapLSP` for Claude; emit an `adapter.Skip` for each LSP server; update the capability matrix (✓→◐/✗), the `README`/`user-guide` matrices, the `docs/`, and the adapter tests. Honest and small-ish. The canonical `lsp/*.toml` schema stays (other agents may use it). **Recommended starting point** unless we want real LSP delivery.
2. **Project LSP as a synthetic Claude plugin** (`.lsp.json` + a generated `plugin.json`, plus enabling it). This is the only way to make Claude actually load an agentsync-managed LSP server — but it's a large change and **fights the read-only-`PluginIngester` invariant** (agentsync deliberately never writes plugin enable-state; see `docs/architecture.md` § PluginIngester (read-only)). Probably out of scope for v1.
3. **Keep as-is and document the limitation.** Rejected by the repo's own rules — a silently-ignored write dressed up as "✓ native" is exactly the silent-drop class CLAUDE.md forbids.

## Acceptance criteria (once a direction is chosen)

- Claude LSP no longer claims `✓ native` unless it lands somewhere Claude Code actually reads (a plugin manifest).
- If downgraded: `settings.json#/lspServers` is no longer written; a Skip is reported; capability matrix + bitmask + tests + docs all agree; the round-trip test is anchored to what Claude actually loads (not the `settings.json` shape).
- Whatever the direction, a test backs the capability claim (per "doc claims are not self-certifying").

## Notes

- OpenCode and Codex already skip LSP for Claude-correct reasons (OpenCode defers it; Codex has no LSP concept), so this is Claude-specific.
- Found during the fan-out audit that produced #56's sibling fix (Claude project MCP → `.mcp.json`, and OpenCode project config → `<root>/opencode.json`).

pranjalbhatia710:
## Summary

Fixes #66.

Claude Code does not load LSP servers from `settings.json#/lspServers`; it only documents LSP servers as plugin manifest components. This changes the Claude adapter to stop advertising/rendering/importing that ignored settings key.

Changes:
- remove `CapLSP` from the Claude adapter
- report canonical Claude LSP servers as explicit skips instead of writing `settings.json#/lspServers`
- stop ingesting stale `settings.json#/lspServers` entries as native Claude LSP
- update tests and docs so the capability matrix matches the on-disk artifact Claude actually reads

## Test plan

```bash
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./...
```

Result: all packages passed locally. I had to set `AGENTSYNC_TEST_IN_CONTAINER=1` because this environment does not have Docker/just available, but the repository test guard otherwise blocks host execution.


pranjalbhatia710:
Rebased this on current `main` and resolved the conflict from the new managed-memory banner work. The branch is clean again and still keeps the Claude LSP behavior scoped to reporting canonical LSP servers as skipped rather than writing/importing `settings.json#/lspServers`.

Validation run after the rebase:

```console
git diff --check upstream/main...HEAD
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./internal/adapter/claude ./internal/cli ./internal/render
# ok github.com/spxrogers/agentsync/internal/adapter/claude 0.039s
# ok github.com/spxrogers/agentsync/internal/cli 39.742s
# ok github.com/spxrogers/agentsync/internal/render 0.072s
```

Note: this runner does not have the repo's `just`/container test harness installed, so I used the project guard env var for the focused package tests rather than claiming a full hermetic `just test` run.

spxrogers:
Reviewing this @pranjalbhatia710

spxrogers:
Thanks for this, @pranjalbhatia710 — solid work. The direction is right (I re-confirmed against Claude's upstream docs: `lspServers` isn't a `settings.json` key, and LSP is documented only as a plugin component), and I especially appreciate that you re-anchored the tests to what Claude *actually* loads instead of the `settings.json` shape. A few things before it can land:

**1. The new LSP skip needs `Kind: adapter.SkipDropped` (currently fails CI on `main`).**
#106 merged after you opened this and made `Skip.Kind` mandatory, enforced by a static AST guard (`TestEverySkipLiteralSetsKind`) plus a behavioral one (`TestEveryAdapterClassifiesSkips`). The new skip in `internal/adapter/claude/render.go` omits it — matching how every other adapter skips LSP (cline/continuedev/cursor/opencode/generic all set `SkipDropped`):

```go
skips = append(skips, adapter.Skip{
    Component: "lsp", Name: l.ID,
    Reason: "Claude Code reads LSP servers only from plugin manifests; agentsync does not synthesize Claude plugins in v1",
    Kind:   adapter.SkipDropped,
})
```

Reproduced against current `main`:

```
$ go test ./internal/adapter/ -run TestEverySkipLiteralSetsKind
skipkind_test.go:126: adapter.Skip literal omits Kind (...): claude/render.go
--- FAIL
```

The green checks here predate #106, so they don't catch it yet.

**2. Please rebase onto current `main`.** The branch is behind by #105–#111 (incl. the #106 guard above); expect a small `CHANGELOG.md [Unreleased]` conflict. After rebasing, please re-run `go test ./internal/adapter/...` (or `just test-fast`).

**3. Three more docs to keep in lockstep** (per CLAUDE.md "keep the docs in sync") — these still claim Claude LSP support:
- `docs/architecture.md` (~L131-133) — frames `CapLSP` as omitted only by OpenCode/Codex/Cursor; after this, *no* adapter has it.
- `website/src/content/docs/help/troubleshooting.mdx` (~L97-100, "LSP beyond Claude") — says LSP "install[s] correctly on Claude."
- `website/src/content/docs/getting-started/existing-configs.mdx` (~L44) — lists **LSP** under the `claude` import selector, which your own test now makes fail.

**4. Could we keep the LSP secret coverage rather than dropping it?** Removing the `claude:lsp:*` cases from `capture_leak_test.go` makes sense — that import path is gone — but LSP fields are still in `walkSecretFields` and are still resolved/re-referenced at apply time, so I'd rather not lose the guard entirely. Could you re-home it as a small `secrets`-package unit test that round-trips a `${secret:…}` through an `LSPServer` directly (via `ReReferenceCanonical` + the residual-leak scan)? That keeps the coverage without needing any adapter to ingest LSP. 🙏

**To land:**
- [ ] Rebase onto `main`
- [ ] Add `Kind: adapter.SkipDropped`
- [ ] Update the three doc spots above
- [ ] Re-home the dropped LSP secret coverage as a `secrets`-package unit test
- [ ] Re-run the adapter tests on the rebased tree

Everything else — the capability bitmask, the matrices, CHANGELOG, and the render/ingest/import test rewrites — looks correct. Ping me once it's rebased and I'll re-review. Thanks again!

pranjalbhatia710:
Thanks for the detailed review — updated and force-pushed the PR branch.

What changed:
- rebased onto current `main` (`e664a63`)
- added `Kind: adapter.SkipDropped` to the Claude LSP skip
- updated the three stale doc spots (`docs/architecture.md`, troubleshooting, existing-configs selector table)
- re-homed the LSP secret coverage into `internal/secrets` with a direct `LSPServer` re-reference + residual cleartext scan test

Validation run locally:

```console
git diff --check upstream/main...HEAD
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./internal/adapter -run TestEverySkipLiteralSetsKind
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./internal/secrets ./internal/adapter/...
```

All passed locally. The first plain `go test ./internal/secrets ./internal/adapter/...` run refused on the host as expected because the repo requires `AGENTSYNC_TEST_IN_CONTAINER=1` for those filesystem-touching packages; rerunning with that env passed.

pranjalbhatia710:
Thanks for the detailed review — updated and pushed `bfffeb3`.

Changes made:
- rebased onto current `main` (`e664a63`) and resolved the `CHANGELOG.md` conflict
- added `Kind: adapter.SkipDropped` to the Claude LSP skip
- updated the three stale docs spots (`docs/architecture.md`, troubleshooting, and `claude` import selector)
- re-homed the LSP secret coverage into `internal/secrets` via `TestReReferenceCanonical_LSPServerSecretFields`
- updated the explain JSON regression expectation now that Claude LSP is explicitly skipped

Validation run:

```console
export PATH=/home/pranjalbhatia/.local/go-toolchains/go/bin:$PATH
go version
# go version go1.26.2 linux/amd64

git diff --check upstream/main...HEAD
git diff --check
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./internal/adapter/... ./internal/secrets ./internal/cli ./internal/render
# ok  	github.com/spxrogers/agentsync/internal/adapter	0.040s
# ?   	github.com/spxrogers/agentsync/internal/adapter/adaptertest	[no test files]
# ok  	github.com/spxrogers/agentsync/internal/adapter/claude	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/cline	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/codex	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/continuedev	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/cursor	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/gemini	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/generic	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/noop	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/opencode	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/roo	(cached)
# ok  	github.com/spxrogers/agentsync/internal/adapter/windsurf	(cached)
# ok  	github.com/spxrogers/agentsync/internal/secrets	(cached)
# ok  	github.com/spxrogers/agentsync/internal/cli	40.297s
# ok  	github.com/spxrogers/agentsync/internal/render	0.061s
```

Note: a plain host `go test` without `AGENTSYNC_TEST_IN_CONTAINER=1` is still blocked by the repo's hermetic-test guard, so I used the same guard env var as before for the focused package set.

pranjalbhatia710:
Thanks for the detailed review — updated in `5611bfb`.

What changed:
- rebased onto current `main`
- added `Kind: adapter.SkipDropped` for Claude LSP skips
- updated the three stale docs spots that still implied Claude LSP support
- re-homed the dropped LSP secret coverage into `internal/secrets` with a direct `LSPServer` `ReReferenceCanonical` regression test + residual leak scan
- updated the explain coverage expectation for LSP-only plugins now that Claude also reports LSP as dropped

Validation run locally after the rebase:

```console
git diff --check upstream/main...HEAD
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./internal/adapter/... ./internal/secrets ./internal/cli ./internal/render
# ok github.com/spxrogers/agentsync/internal/adapter ...
# ok github.com/spxrogers/agentsync/internal/secrets
# ok github.com/spxrogers/agentsync/internal/cli 40.151s
# ok github.com/spxrogers/agentsync/internal/render
```

spxrogers:
merged! thank you @pranjalbhatia710! 🚀

spxrogers:
@pranjalbhatia710 v0.9.0 has been released with this change 

https://github.com/spxrogers/agentsync/releases/tag/v0.9.0

PR #91, opened 2026-06-17
feat(memory): prepend an agentsync-managed banner to rendered memory files

spxrogers:
## What

Every rendered memory file (`CLAUDE.md`, `AGENTS.md`, `GEMINI.md`, the breadth-tier rules files, …) now carries a short banner naming the file and pointing edits back at the canonical source:

```markdown
<!-- agentsync:managed -->
> **Managed by [agentsync](https://agentsync.cc) — do not edit `CLAUDE.md` directly.**
> To change it, edit `.agentsync/memory/AGENTS.md` (or the relevant
> `.agentsync/memory/fragments/*.md` fragment) and run `agentsync apply`.
> Direct edits here are reported as drift and overwritten on the next apply.
<!-- /agentsync:managed -->
```

The goal: an enforced agentsync-awareness so an agent (or teammate) editing the native file is nudged toward the canonical source instead of making a rogue direct edit.

**Settings (chosen up front):** on by default (`[memory] banner = false` to opt out), prepended at the top, filename-aware wording.

## How it stays safe in the bidirectional model

The core hazard was a self-injected banner leaking back into the canonical `memory/AGENTS.md` and compounding on every apply. The design mirrors the existing fragment-marker round-trip:

- **One render chokepoint** — all 31 adapters now assemble memory through `source.RenderManagedMemory` (wraps `ExpandMemoryImports`), so the banner is byte-identical across agents. Windsurf's activation frontmatter stays at byte 0; the banner sits after it.
- **Stripped on the way back** — wrapped in reversible `<!-- agentsync:managed -->` markers and removed by `source.StripManagedBanner` in each adapter's ingest (the inverse of render, like Windsurf's frontmatter strip) **and** as a backstop at the `import`/`reconcile` write-back funnels (reconcile reads the dest raw, no ingest). So it never enters the canonical source and never compounds.
- **No false drift** — the banner text is static (only the filename varies), so it hashes identically across re-render / state / disk; an untouched file still classifies `InSync`.
- **Config** — on by default; `[memory] banner = false` in `agentsync.toml` opts out. The project overlay inherits the effective user setting unless it sets its own.

## Behavior change

The first `apply` after this lands rewrites managed memory files to add the banner — a one-time `pending` in `status`, then stable.

## Tests

- Source-level units + an **artifact-anchored round-trip** (on-disk fixture with `@import` + fragment → render → strip → collapse → write → byte-identical, with the banner present in neither canonical file).
- Adapter ingest round-trips now exercise banner-on render → ingest-strip; render exact-content assertions updated to strip the banner.
- Full suite green: 26 internal packages, e2e, BDD, `golangci-lint` 0 issues, gofumpt/`go mod tidy` clean.

## Docs (updated in the same commit)

`docs/concepts.md`, `docs/architecture.md`, `docs/components.md`, `docs/user-guide.md`, `CHANGELOG.md`, and the website memory guide + configuration reference.

## Note on dogfooding

This repo is agentsync-managed, so its own `CLAUDE.md`/`AGENTS.md` won't show the banner until someone runs `agentsync apply --scope project` to re-render them — deliberately left as a separate, explicit step rather than folded into this PR.

https://claude.ai/code/session_01Xvvan2aDifmSpjhAwic1yg

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xvvan2aDifmSpjhAwic1yg)_

### DIFF
--- a/internal/source/memory.go
+++ b/internal/source/memory.go
@@ -115,26 +115,57 @@ func expandMemoryImports(body string, fragments map[string]string, visiting map[
 // blockquote) so an agent reading the rendered CLAUDE.md/AGENTS.md sees the
 // notice; the markers themselves are HTML comments the agent ignores.
 //
-// The token "agentsync:managed" is RESERVED: if the canonical memory already
-// contains it, RenderManagedMemory skips injection (collision guard, mirroring
-// ExpandMemoryImports' marker-collision handling) so a second banner can't nest
-// and StripManagedBanner can't mis-bound a user's literal.
+// The token "agentsync:managed" is RESERVED: canonical memory (the AGENTS.md body
+// or any fragment) must not contain it, and checkReservedMarkers (called from
+// loadMemory and WriteMemory) rejects a canonical that does — with a message
+// telling the user to remove/rename the offending content — rather than letting it
+// collide with the banner's markers. StripManagedBanner additionally anchors on
+// the banner's lead line (managedBannerLead), not the bare markers, so even a
+// hand-edited native file that happens to carry a user-authored
+// `<!-- agentsync:managed -->` block has its content preserved (the write-back
+// then surfaces the reserved-token error) — it is never silently deleted.
 const managedMarkerToken = "agentsync:managed"
 
 const (
 	managedStartMarker = "<!-- agentsync:managed -->"
 	managedEndMarker   = "<!-- /agentsync:managed -->"
+	// managedBannerLead is the fixed opening of the banner body (everything before
+	// the per-file name). StripManagedBanner anchors on it so it removes ONLY
+	// agentsync's own banner; managedBanner builds from the same constant so the
+	// two halves can never drift.
+	managedBannerLead = "> **Managed by [agentsync](https://agentsync.cc)"
 )
 
-// managedBannerRe matches a whole injected banner block — the markers, the body
-// between them, and the blank line that separates it from the memory content —
-// anywhere in the file. `(?s)` lets `.` span lines; `.*?` is non-greedy so each
-// block is bounded minimally; `\n{0,2}` consumes the trailing separator we add
-// (and tolerates a hand-removed blank line). ReplaceAll over it makes
-// StripManagedBanner idempotent and robust to a (theoretical) double banner.
+// managedBannerRe matches a whole injected banner block — the markers, the banner
+// body between them, and the blank line that separates it from the memory content.
+// It anchors on managedBannerLead immediately after the start marker, so it only
+// matches agentsync's own banner, never an arbitrary user-authored
+// `<!-- agentsync:managed -->` block. `(?s)` lets `.` span lines; `.*?` is
+// non-greedy so the block is bounded minimally; the `\r?\n` / `(?:\r?\n){0,2}`
+// tolerate CRLF line endings and consume the trailing separator the renderer adds
+// (and a hand-removed blank line). ReplaceAll makes StripManagedBanner idempotent.
 var managedBannerRe = regexp.MustCompile(
-	`(?s)` + regexp.QuoteMeta(managedStartMarker) + `.*?` + regexp.QuoteMeta(managedEndMarker) + `\n{0,2}`,
+	`(?s)` + regexp.QuoteMeta(managedStartMarker) + `\r?\n` + regexp.QuoteMeta(managedBannerLead) +
+		`.*?` + regexp.QuoteMeta(managedEndMarker) + `(?:\r?\n){0,2}`,
 )
+
+// checkReservedMarkers rejects canonical memory that contains the reserved
+// managed-banner token. The token is owned by the managed-file banner
+// (RenderManagedMemory); a body or fragment carrying it would collide with the
+// banner's reversible markers, so agentsync errors and asks the user to fix it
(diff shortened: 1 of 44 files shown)


internal/source/memory.go
```diff
@@ -107,6 +107,89 @@ func expandMemoryImports(body string, fragments map[string]string, visiting map[
...
+// The token "agentsync:managed" is RESERVED: if the canonical memory already
+// contains it, RenderManagedMemory skips injection (collision guard, mirroring
+// ExpandMemoryImports' marker-collision handling) so a second banner can't nest
+// and StripManagedBanner can't mis-bound a user's literal.
+const managedMarkerToken = "agentsync:managed"
+
+const (
+	managedStartMarker = "<!-- agentsync:managed -->"
+	managedEndMarker   = "<!-- /agentsync:managed -->"
+)
```
spxrogers, line 146: Since this feature is implemented via a quasi-fragment (good!), we'll need some sort of checks that that no agentsync/memory/fragment/* name conflicts and would create a marker ID of `agentsync:managed`. 

If a user has a fragment that would generated this marker, it needs to be an error and prompt the user to rename their custom memory fragment.

spxrogers: this _shouldn't_ happen with today's logic given the fragment markers are the format:

```
<!-- agentsync:fragment filename.md -->
<!-- /agentsync:fragment filename.md -->
``` 

but if sensical, we should harden to prevent an unintended future regression.

spxrogers: On that vein; we should update this marker to be:

```golang
const (
	managedStartMarker = "<!-- agentsync:managed memory-banner -->"
	managedEndMarker   = "<!-- /agentsync:managed memory-banner -->"
)
```

so that _**if**_ agentsync starts having more managed markers in the future, each managed marker has its own Identifier for stable bidirectional parsing.

spxrogers: Done in ac5526b. `agentsync:managed` is now a **reserved marker**: `checkReservedMarkers` (called from `loadMemory` *and* `WriteMemory`) rejects, with a clear error, any canonical memory whose `AGENTS.md` body or a fragment contains it — so it surfaces both when authoring (any command that loads) and at capture (import/reconcile refuse to persist it), prompting the user to remove/rename rather than silently degrading.

A reviewer also caught the mirror-image hazard: the old strip removed *any* `<!-- agentsync:managed -->…<!-- /agentsync:managed -->` block, so a user-authored block in a native file would've been silently deleted on capture. `StripManagedBanner` now anchors on the banner's own signature line (`managedBannerLead`, shared with the builder so the two can't drift) — it strips only agentsync's banner and preserves a user block verbatim. Both directions are break-verified and covered by tests (`TestLoad_RejectsReservedMarker`, `TestWriteMemory_RejectsReservedMarker`, `TestStripManagedBanner_PreservesUserAuthoredBlock`).

One judgment call worth flagging: I made it a hard **load-time** error (every command that loads the canonical fails until it's removed), matching "it needs to be an error." If you'd prefer it confined to `verify`/`doctor` (so read-only commands like `status` still run), say the word and I'll move it.

---

spxrogers: Done in 73011e1 — sorry I jumped on the first comment before reading the whole thread.

The marker is now namespaced exactly as you specced:

```go
const managedMarkerNamespace = "agentsync:managed"
const (
	managedStartMarker = "<!-- " + managedMarkerNamespace + " memory-banner -->"
	managedEndMarker   = "<!-- /" + managedMarkerNamespace + " memory-banner -->"
)
```

`agentsync:managed` is the reserved *namespace* and `memory-banner` is this marker's identifier, so future managed markers each get their own ID. `checkReservedMarkers` reserves the whole namespace (rejects any `agentsync:managed` in canonical memory at load **and** capture).

One thing I went a step further on, prompted by the parallel review: rather than parsing/stripping purely by the markers, `StripManagedBanner` matches agentsync's **full rendered banner** (derived from one shared template, so render and strip can't drift). Reason — a hand-edited *native* file isn't subject to the canonical load check, so a user block that reused the markers (even one opening with the exact banner lead) would otherwise be silently deleted on import. Matching the whole banner means any block that differs in any line is preserved and then rejected *loudly* by `checkReservedMarkers`, never silently dropped. New test covers bare markers, exact-markers-different-body, and banner-like-opener cases, plus a CLI test that `import claude:memory` of such a native block surfaces the reserved-marker error.

If you'd prefer pure marker-ID-based bidirectional parsing (simpler, and the namespacing + reserved-check already makes collisions vanishingly unlikely) over the full-banner match, happy to swap — just let me know.

Still your call on the earlier open question too: hard **load-time** error vs. confining it to `verify`/`doctor`.

---

spxrogers: Hard load time error.

spxrogers: 👍 That's the current behavior — no change needed. `checkReservedMarkers` is called inside `loadMemory` (and `WriteMemory`), so a reserved `agentsync:managed` marker in canonical memory hard-errors every command that loads the source until it's removed. Covered by `TestLoad_RejectsReservedMarker`. Leaving the full-banner-match strip in place too (you didn't ask to swap it back to pure marker-ID parsing). I think this thread's resolved.

---

Issue #178, opened 2026-06-30
EPIC — agentsync v1.0 Remediation (54 issues, 4 waves)


spxrogers:
[!NOTE]
> **Update (post-creation): pre-existing issue #66 was resolved after this epic was filed.**
> #66 (Claude LSP written to `settings.json#/lspServers`, which Claude Code reads only from plugin manifests) — flagged at hand-off as a candidate to fold into Wave 1/2 — was **fixed by PR #73** (Claude LSP downgraded to an `adapter.Skip`; `CapLSP` dropped; matrix/docs/tests updated) plus follow-up **PR #180** (re-homed the render-pipeline sibling-section/orphan-cleanup tests onto Codex). **Do not add #66 to the plan.**
>
> Side effects on the issues below (each carries its own "Update since filed" note): **#159** and **#153** lose their Claude-LSP sub-scope (hooks-only now); **#124 / #151 / #177** have minor re-scoping and `file:line` drift; **#162 / #167** are ~unaffected. No waves, dependencies, or issue count change.
>
> **Issues filed after this epic (not among the 54):** **#181** (re-add Claude LSP via plugin manifests — the sequel to #73’s `Skip`), **#182** (first-class canonical `Rule` component — a *feature*, out of remediation scope), and **#188** (re-enable the Chocolatey release once it clears community moderation — the pipe was commented out in **#189**; intersects release issues #138/#173). The project is at **v0.10.1**; this epic targets the **v1.0** milestone. None of these change the 54-issue plan.

---

# EPIC — agentsync v1.0 Remediation (54 issues, 4 waves)

This epic tracks the full remediation plan derived from the **agentsync v1.0 release review** (blockers → highs → mediums/lows → cross-cutting hardening & QA). It bundles **54 issues** across **4 dependency-sequenced waves**, authored to be self-contained for a Claude Ultracode execution session.

- **Items captured:** 255 review findings (248 actionable) → 54 issues. 7 items were upstream-verified as resolved/refuted and intentionally excluded (see the review's §F).
- **Two blockers gate the release:** #124 (Claude hook model) and #128 (revert deletes untracked files).
- **Execution model:** waves 1/3/4 are internally fully parallel; wave 2's git cluster serializes behind #128. Recommended concurrency: 8 lanes.

## Issue legend & dependency tree

[...]

Issue #150, opened 2026-06-30
docs(matrix): correct the Codex subagent `name` "carries over" claim to reflect ingest fidelity


spxrogers:
## 1. Metadata

- **Severity:** high
- **Area:** docs
- **Wave:** 3
- **Themes:** C2 (doc/code drift on contract surfaces)
- **Depends on:** `#144` — *fix(codex): re-populate subagent frontmatter name on ingest and detect colliding effective names* (the code fix this issue documents)
- **Can run in parallel with:** Nothing that touches `docs/capability-matrix.md` lines 208–220 (single matrix cell). This issue is **blocked-by #144** and must land in or after the same change, because the restated matrix claim is only true once the ingest behavior is fixed (or, if #144 chose to surface the loss instead of fixing it, the matrix must describe whatever #144 actually shipped). Do not start the prose edit until #144's final behavior is known.
- **Labels:** `severity:high`, `area:docs`, `theme:C2`, `wave:3`

## 2. Summary

The Codex row of the capability matrix (`docs/capability-matrix.md`, "What each ◐ loses" → **Codex → Subagent**) asserts that a subagent's `name` frontmatter **"carries over."** That is false on the *ingest* (native → canonical) side: Codex ingest reconstructs the canonical name from the **filename** and never reads the TOML `name` field back into frontmatter, so a Codex agent whose `name` field intentionally diverges from its file stem has that value silently erased on the next `import`/`reconcile`. The matrix is a contract page, so this claim actively hides the silent-drop bug (the #144 / blocker B#3 defect) from a user consulting it. **Done** = the Codex Subagent matrix entry truthfully describes the round-trip fidelity of `name` as actually shipped by #144, with no remaining "carries over" wording that overstates ingest behavior, and `docs/capability-matrix.md` stays in sync with its website-generated mirror.

## 3. Background & context (for an agent with zero context)

agentsync is a single-machine Go CLI that centrally manages AI coding-agent configs. The user keeps a canonical config (`source.Canonical`, TOML + markdown) and `agentsync apply` *renders* it into each agent's native files; native edits are detected as drift and merged back via `reconcile`/`import` (the *ingest* direction). Codex is one of the nine deep adapters (`internal/adapter/codex/`).

**The contract page.** `docs/capability-matrix.md` is the per-agent support matrix. Its "What each ◐ loses" section is the authoritative, human-readable statement of exactly what each adapter *projects* vs *drops*. Two CLAUDE.md invariants govern it:

[...]

Issue #149, opened 2026-06-30
fix(git): re-check for a nested foreign repo before Restore and on the commit path


spxrogers:
## 1. Metadata

- **Severity:** medium
- **Area:** git (`internal/git`, `internal/cli` git-backup/revert/doctor)
- **Wave:** 2
- **Themes:** C3
- **Depends on (blocked-by):**
  - **#128** — `fix(git): stop agentsync revert from deleting the user&#39;s untracked files in Restore` (shared `Restore`/`HardReset` mechanism — land that first so the nested-repo guard sits on top of the corrected Restore semantics).
- **Can run in parallel with:** #146 / #143 only via coordination on the revert path (all three touch `internal/cli/revert.go`); **not** parallel-safe against #128 (same Restore/HardReset code).
- **Labels:** `severity:medium`, `area:git`, `theme:C3`, `wave:2`

---

## 2. Summary

`agentsync`&#39;s destination git-backup machinery guards against wrapping a foreign git repo **only at init time** (`HasNestedRepoBelow`). Once a directory is agentsync-owned, every later `apply` (commit path) and every `revert` (Restore → hard reset) operates over the whole work tree with **no re-check**, so a foreign repo a user clones under an owned root *after* init (e.g. `~/.claude/some-plugin/.git`) gets its checked-out files clobbered by a hard reset to the agentsync checkpoint tree. The nesting scan also uses `filepath.WalkDir`, which does not follow symlinks, so a symlinked foreign repo defeats the init-time guard too. Finally, `doctor` reports a clean `untracked` for a root that `apply` will actually refuse to init, so its report lies about what apply does. **Done** = the commit and revert paths re-probe for a nested foreign repo before any destructive write and refuse/warn-and-skip; symlinked nested repos are either detected or explicitly documented out of scope; and `doctor` downgrades a `StateUntracked` root that contains a nested `.git` to a warn so its report matches apply.

---

## 3. Background &amp; context (for an agent with zero context)

[...]

Issue #148, opened 2026-06-30
fix(opencode): add ingest ownership filter and preserve primary/all agent mode across round-trip


spxrogers:
## 1. Metadata

- **Severity:** high
- **Area:** adapters (OpenCode deep adapter)
- **Wave:** 2
- **Themes:** C1 (impoverished-subset / silent-drop), C2 (doc/code drift on contract surfaces)
- **Depends on:** none (not blocked)
- **Can run in parallel with:** #125 (`fix(opencode): preserve unmodeled subagent/command frontmatter`). #125 touches `internal/adapter/opencode/subagent.go` and `command.go` (the **render** side); this issue touches `internal/adapter/opencode/ingest.go` (the **ingest** side) plus the `mode` handling in `subagent.go`. Edits are disjoint but the `subagent.go` render-frontmatter shape is shared — coordinate to avoid stepping on the same `out := map[string]any{}` block.
- **Labels:** `severity:high`, `area:adapters`, `theme:C1`, `wave:2`

## 2. Summary

OpenCode&#39;s `Ingest` (the inverse of `Render`) over-captures the **entire** native `agents/` and `commands/` directories — every `.md` file, including agents and slash-commands the user authored by hand that agentsync never wrote — and it unconditionally **deletes the `mode` frontmatter key** on every captured agent. Because the render path always re-emits `mode: subagent`, a native OpenCode agent declared `mode: primary` or `mode: all` is silently **demoted to `subagent`** on the next `import`/`reconcile` → `apply` round-trip, and unmanaged user files get pulled into the canonical source. &#34;Done&#34; means: (a) ingest only captures files agentsync owns (an ownership filter), and (b) the agent `mode` (`primary` / `all` / `subagent`) survives the round-trip — both proven by a **spec-complete on-disk fixture**, not a model round-trip.

## 3. Background &amp; context (for an agent with zero context)

agentsync is a single-machine Go CLI that centrally manages AI coding-agent configs. The user keeps a canonical config (TOML + markdown) in `~/.agentsync/`; `agentsync apply` **renders** it into each agent&#39;s native config; native edits are detected as drift and merged back via **`reconcile`/`import`**, which call each adapter&#39;s **`Ingest`** (the inverse of `Render`). The relevant package here is `internal/adapter/opencode/` — a *deep* adapter implementing the `Adapter` interface.

[...]

PR #194, opened 2026-07-19
epic #178 — Wave 2: carried-over correctness highs + git-backup/revert data-loss & safety fixes

### LINKED ISSUE #140: fix(cli): preserve a plugin's agents allowlist (and Update/Disabled) across install and disable→enable
## 1. Metadata

- **Severity:** high
- **Area:** cli
- **Wave:** 2
- **Themes:** credential-fan-out / secret-adjacent data loss; &#34;surface the change, don&#39;t report blind success&#34;
- **Depends on:** none (no blocking issues)
- **Can run in parallel with:** all other Wave-2 work — this issue touches only `internal/cli/plugin.go` and its tests, independent of every other Wave-2 item.
- **Labels:** `severity:high`, `area:cli`, `wave:2`

## 2. Summary

`agentsync plugin install` and `agentsync plugin disable` → `agentsync plugin enable` silently **widen or reset** a plugin&#39;s `agents` allowlist (and, for install, blow away `Update`/`Disabled`), so credential-bearing MCP servers / skills / hooks fan out to agents the user *intentionally excluded*. Both operations report blanket success and never tell the user their targeting changed. &#34;Done&#34; means: `install` (re-install of an already-registered plugin) and `disable→enable` **preserve** the existing `Agents`/`Update`/`Disabled` fields instead of overwriting them; any change the operation does make to the allowlist is surfaced to the user rather than hidden behind a cheerful success line; and a regression test proves a narrowed allowlist survives both code paths.

## 3. Background &amp; context (for an agent with zero context)

agentsync is a single-machine Go CLI that centrally manages AI coding-agent configs. The user keeps a canonical config under `~/.agentsync/`; `agentsync apply` renders it into each agent&#39;s native config.

**Plugins.** A plugin is a bundle (MCP servers, skills, subagents, commands, hooks, LSP servers) fetched from a *marketplace*. Installing a plugin writes a small TOML file at `~/.agentsync/plugins/.toml`. At apply time, `marketplace.LoadProjected` fans the plugin&#39;s components out to the agents named in the plugin&#39;s `agents` allowlist. The canonical model for that TOML is `source.PluginSpec` (`internal/source/schema.go:158-173`):

```go
type PluginSpec struct {
	ID          untrusted.Text `toml:&#34;id&#34;`
	Version     untrusted.Text `toml:&#34;version,omitempty&#34;`
	ManifestSHA string         `toml:&#34;manifest_sha,omitempty&#34;`
	Update      string         `toml:&#34;update,omitempty&#34;` // pinned | track | manual
	Agents      []string       `toml:&#34;agents,omitempty&#34;`
	// Disabled, when true, suppresses the plugin&#39;s projection during
	// marketplace.LoadProjected. `agentsync plugin disable ` sets this.
	Disabled bool `toml:&#34;disabled,omitempty&#34;`
}
```

The CLI&#39;s own copy of this shape is `pluginTOMLSpec` in `internal/cli/plugin.go:42-52`, with the same `Agents []string`, `Update string`, `Disabled bool` fields.

**The `agents` allowlist is user-facing, documented, security-relevant config.** The published docs tell users to narrow fan-out with it (`website/src/content/docs/guides/plugins.mdx:69-82`):

&gt; Limit a plugin to specific agents with an `agents = [...]` allowlist in its TOML file:
&gt; ```toml
&gt; id     = &#34;atlassian@anthropic&#34;
&gt; agents = [&#34;claude&#34;]   # only fan this plugin out to Claude
&gt; ```

So a user who set `agents = [&#34;claude&#34;]` did so deliberately: the plugin ships credentials (MCP servers carry `${secret:…}` headers/env; hooks carry commands), and they want those reaching Claude *only*. Widening or resetting that allowlist back to `[&#34;*&#34;]` fans those credentials out to every enabled agent — exactly the blast radius the user opted out of. That ties this directly to CLAUDE.md&#39;s secret-handling concern: **the dangerous bug class is a resolved cleartext secret reaching somewhere the user did not intend.** Quietly re-broadening an allowlist is a config-side version of that same hazard.

CLAUDE.md also makes two doctrines load-bearing here:

- **Models must stay faithful to their on-disk artifacts — non-negotiable.** &#34;The on-disk artifact … is the source of truth, never the struct that parses it … Capture it or acknowledge it — never drop it silently. … A silent drop is a bug, full stop.&#34; Overwriting `Agents`/`Update`/`Disabled` on an existing plugin TOML is a silent drop of user-authored on-disk state.
- The CLI convention already established in this very file: a no-op or a destructive action should not be &#34;presented as success.&#34; (See `pluginRemoveRun` at `plugin.go:319-324`, which deliberately errors on a typo&#39;d id &#34;rather than a cheerful &#39;removed plugin X&#39; no-op.&#34;)

**Where the code lives.** Everything is in `internal/cli/plugin.go` (package `cli`). The relevant functions are `pluginInstallRun` → `installPluginInto` (install), `pluginEnableRun` (enable), and `pluginDisableRun` (disable). The shared TOML read/write helpers are `readPluginTOML` (`plugin.go:553-564`) and `iox.AtomicWrite`. Tests live in `internal/cli/plugin_test.go` (package `cli_test`), driven through the `runCLI`/`mustRun` helpers against a `t.TempDir()` `AGENTSYNC_TARGET_ROOT`.

## 4. Problem

There are three distinct data-loss sites, all in `internal/cli/plugin.go`. Each overwrites user-authored on-disk allowlist/lifecycle state and reports success.

### 4a. `disable` resets `Agents` to empty (`plugin.go:287-288`)

`pluginDisableRun` reads the existing TOML, then:

```go
existing.Plugin.Disabled = true
existing.Plugin.Agents = []string{}
```

It deliberately erases the allowlist. A user with `agents = [&#34;claude&#34;]` who runs `plugin disable demo` loses that targeting — the field is now empty on disk.

### 4b. `enable` re-broadens the (now-empty) `Agents` to `[&#34;*&#34;]` (`plugin.go:249-252`)

`pluginEnableRun` does:

```go
existing.Plugin.Disabled = false
if len(existing.Plugin.Agents) == 0 {
	existing.Plugin.Agents = []string{&#34;*&#34;}
}
```

Because `disable` (4a) emptied `Agents`, the very next `enable` sees `len == 0` and &#34;helpfully&#34; widens it to `[&#34;*&#34;]` — every agent. So the round-trip `disable` → `enable` turns `agents = [&#34;claude&#34;]` into `agents = [&#34;*&#34;]`. The two bugs compose: the narrowed allowlist the user set is **silently widened to all agents**, fanning the plugin&#39;s credential-bearing components out to agents they excluded. Both commands print success (`&#34;disabled plugin demo&#34;` / `&#34;enabled plugin demo&#34;`).

### 4c. `install` (re-install) hard-resets `Agents`/`Update`/`Disabled` (`plugin.go:95-148`)

`installPluginInto` always constructs a fresh spec and writes it, with no read-modify-write of an existing TOML:

```go
spec := pluginTOMLSpec{
	ID:          untrusted.Wrap(id + &#34;@&#34; + resolveMarketplaceName(mpName)),
	Version:     mpEntry.Version,
	ManifestSHA: manifestSHA,
	Update:      &#34;track&#34;,
	Agents:      []string{&#34;*&#34;},
}
...
if err := iox.AtomicWrite(pluginPath, data, 0o644); err != nil {
```

`installPluginInto` writes to `plugins/.toml` unconditionally. So **re-installing** an already-registered plugin (e.g. to refresh after a marketplace update, or because the user re-ran the install command) hard-resets `Agents` back to `[&#34;*&#34;]`, `Update` back to `&#34;track&#34;`, and drops any `Disabled = true` — wiping a narrowed allowlist, a `pinned`/`manual` update mode, and a deliberate disabled state, all at once, while printing `&#34;installed plugin demo (…)&#34;`. (Note: `installPluginInto` is shared by `plugin install` and `import`; the doc comment at `plugin.go:90-94` says both must &#34;produce byte-identical canonical artifacts,&#34; so any fix here must keep that property — see §5.)

### Why this is high severity

Every site silently broadens or resets a user&#39;s deliberate, security-relevant fan-out scope and reports unqualified success. The widened scope ships credential-bearing MCP/skills/hooks to excluded agents on the next `apply`. It is the config-targeting analogue of the secret-fan-out hazard CLAUDE.md is built to prevent, and a textbook &#34;silent drop of on-disk state&#34; per the models-faithful doctrine.

## 5. Required changes

All edits are in `internal/cli/plugin.go`. Respect the project invariants: this is **not** a dest→source write (no `capture.Capture` needed — these are direct edits to a canonical `plugins/*.toml`), no new secret-bearing field is introduced (so `walkSecretFields` is untouched), and no adapter `Render`/`DestWriter` path is involved.

1. **`pluginDisableRun` (`plugin.go:275-298`) — stop erasing `Agents`.** Remove `existing.Plugin.Agents = []string{}`. Set only `existing.Plugin.Disabled = true` and write back. The allowlist (and `Update`) must survive disable untouched. The `disabled = true` bit alone already suppresses projection (`source.PluginSpec.Disabled` doc, `schema.go:167-172`), so emptying `Agents` was never necessary.

2. **`pluginEnableRun` (`plugin.go:237-262`) — stop re-broadening.** Remove the `if len(existing.Plugin.Agents) == 0 { existing.Plugin.Agents = []string{&#34;*&#34;} }` block. Set only `existing.Plugin.Disabled = false` and write back. With (1) in place, a previously-narrowed allowlist is still present at enable time; do not touch it. (A plugin legitimately authored with an empty/omitted `agents` already means &#34;all agents&#34; downstream — re-materializing `[&#34;*&#34;]` is both unnecessary and the exact mechanism that destroyed the narrowed value.)

3. **`installPluginInto` (`plugin.go:95-148`) — preserve existing lifecycle fields on re-install.** Before constructing the new spec, attempt to read an existing `plugins/.toml` (it may not exist on a first install — a not-exist read is fine, fall through to defaults). If it exists, **carry forward** the user-authored fields rather than resetting them:
   - `Agents`: keep the existing value if set; default to `[&#34;*&#34;]` only when there is no existing TOML (genuine first install) or the existing `Agents` is unset.
   - `Update`: keep the existing value if non-empty; default to `&#34;track&#34;` only on first install.
   - `Disabled`: preserve the existing bit (re-installing a disabled plugin should not silently re-enable it).
   - `ID`/`Version`/`ManifestSHA`: continue to come from the fresh fetch (these are the things install legitimately refreshes).

   Keep `installPluginInto` deterministic and side-effect-shaped exactly as today so the shared-with-`import` &#34;byte-identical canonical artifacts&#34; contract (`plugin.go:90-94`) still holds: a **first** install with no prior TOML must produce the same bytes it does today (`Agents=[&#34;*&#34;]`, `Update=&#34;track&#34;`, `Disabled=false`). Only the **re-install over an existing TOML** path changes.

4. **Surface the change instead of reporting blind success.** Per the existing CLI convention (the no-cheerful-no-op rule at `plugin.go:319-324`) and CLAUDE.md&#39;s &#34;capture it or acknowledge it&#34;:
   - For `install`, when an existing TOML&#39;s preserved fields differ from the install defaults (e.g. a narrowed `Agents`, a non-`track` `Update`, or `Disabled=true`), note it in the status line so the user knows the existing targeting/lifecycle was kept (e.g. append `(kept agents=[…], update=…, disabled=…)` or an equivalent concise message). Do not silently default-overwrite.
   - For `disable`/`enable`, since the fix now *preserves* the allowlist, no extra surfacing is strictly required — but the success line must not imply the allowlist was reset. (It currently says only `&#34;disabled plugin demo&#34;`/`&#34;enabled plugin demo&#34;`, which is fine once the reset is gone.)

5. **Keep `Version`/`id` print-sanitization intact.** `spec.Version` is `untrusted.Text` printed via `%s` for ANSI-escape safety (`plugin.go:81-86`); don&#39;t regress that when adding any new status output — route untrusted values through `%s`/`ui.Sanitize` as the surrounding code already does.

## 6. Acceptance criteria

- [ ] `plugin disable ` sets `disabled = true` and **leaves `agents` (and `update`) unchanged** on disk.
- [ ] `plugin enable ` sets `disabled = false` and **leaves `agents` unchanged**; it never re-materializes `agents = [&#34;*&#34;]` over a previously-narrowed (now non-empty) allowlist.
- [ ] `disable` → `enable` round-trip on a plugin with `agents = [&#34;claude&#34;]` yields `agents = [&#34;claude&#34;]` again (not `[&#34;*&#34;]`).
- [ ] Re-running `plugin install ` on a plugin whose TOML has `agents = [&#34;claude&#34;]`, `update = &#34;pinned&#34;`, `disabled = true` **preserves** all three (does not reset to `[&#34;*&#34;]`/`&#34;track&#34;`/`false`).
- [ ] A genuine **first** install (no prior `plugins/.toml`) still writes `agents = [&#34;*&#34;]`, `update = &#34;track&#34;`, `disabled` absent — byte-identical to today, so the `install`/`import` shared-artifact contract (`plugin.go:90-94`) holds.
- [ ] When `install` preserves a non-default allowlist/update/disabled state, the change is **surfaced** in the command output (not hidden behind an unqualified success line).
- [ ] No new secret-bearing field is introduced; `walkSecretFields` and `TestNewSecretFieldGuard` are untouched. No new dest→source write bypassing `capture.Capture` is added (these are canonical-source edits, not write-backs).
- [ ] `just test` (in-container) passes, including the new/updated tests below; `just lint` is clean and any files it rewrites are re-staged.

## 7. Tests to add / update

All in `internal/cli/plugin_test.go` (package `cli_test`), following project conventions: stdlib `testing` only, table-driven with a `name` field + `t.Run` where there are multiple cases, `t.TempDir()` + `AGENTSYNC_TARGET_ROOT` env, driven through the existing `runCLI`/`mustRun` helpers and `makeLocalMarketplace`. These are FS-touching CLI tests; they run through the same harness as the existing plugin tests (which already exercise the container path), so they inherit the container guard — do not add a separate `os.UserHomeDir` access.

**Anchor assertions to the on-disk artifact, not a re-parsed model.** Per CLAUDE.md&#39;s models-faithful doctrine, the round-trip test must start from a real `plugins/.toml` written with a narrowed allowlist and assert the **on-disk file** still carries it after the operation — read the file back and parse it (or assert on its bytes), do not assert on an in-memory struct that never touched disk.

1. **`TestPlugin_DisableEnablePreservesAgentsAllowlist`** — install `demo@test-mp`; rewrite `plugins/demo.toml` so `[plugin].agents = [&#34;claude&#34;]` (and `update = &#34;pinned&#34;`); run `plugin disable demo` then `plugin enable demo`; read `plugins/demo.toml` back from disk and assert `agents == [&#34;claude&#34;]` (NOT `[&#34;*&#34;]`, NOT empty), `update == &#34;pinned&#34;`, and `disabled` is false/absent. Also assert an intermediate read after `disable` shows `agents == [&#34;claude&#34;]` and `disabled = true` (i.e. disable did not empty it).

2. **`TestPlugin_ReinstallPreservesAgentsUpdateDisabled`** — install `demo@test-mp`; rewrite `plugins/demo.toml` to `agents = [&#34;claude&#34;]`, `update = &#34;manual&#34;`, `disabled = true`; re-run `plugin install demo@test-mp`; read the file back and assert all three are preserved (`agents == [&#34;claude&#34;]`, `update == &#34;manual&#34;`, `disabled == true`), while `manifest_sha`/`version` reflect the fresh fetch. Assert the command output surfaces that the existing allowlist/lifecycle was kept (acceptance criterion §6 bullet 6).

3. **`TestPlugin_FirstInstallDefaults`** (or extend the existing `TestPlugin_InstallFromLocalMarketplace`) — on a clean home with no prior TOML, install `demo@test-mp` and assert the written `plugins/demo.toml` has `agents = [&#34;*&#34;]`, `update = &#34;track&#34;`, and `disabled` absent — pinning the byte-identical-to-today first-install behavior so the `install`/`import` shared-artifact contract is not regressed.

4. Keep the existing `TestPlugin_EnableDisable` and `TestPlugin_DisableSuppressesProjectionAtApply` green; if `TestPlugin_EnableDisable` implicitly relied on the reset behavior, update its assertions (it currently only checks the `enabled`/`disabled` *status* word, which the fix does not change).

## 8. Docs to update

Per the CLAUDE.md docs-sync contract, update in the **same change**:

- `website/src/content/docs/guides/plugins.mdx` (the &#34;Controlling per-plugin fan-out&#34; + &#34;Plugin lifecycle&#34; sections, ~lines 69-93): add a sentence making explicit that `disable`/`enable` and re-`install` **preserve** the `agents` allowlist (and `update`/`disabled`) — i.e. lifecycle commands don&#39;t silently re-broaden a narrowed plugin. This is the user-facing contract page for the allowlist; today it tells users to rely on `agents = […]` without warning that lifecycle commands used to reset it.
- `CHANGELOG.md` under `[Unreleased]`: a `Fixed` entry — `plugin install`/`disable`/`enable` no longer silently widen or reset a plugin&#39;s `agents` allowlist (and `update`/`disabled`).
- `website/src/content/docs/reference/cli.mdx` and `docs/user-guide.md`: check the `plugin install`/`enable`/`disable` command descriptions; if either claims or implies an allowlist reset/default behavior, correct it. If neither says anything about the allowlist on these subcommands, no change is needed there beyond the plugins-guide note — state that explicitly in the PR.
- **Capability matrix:** no change — this is a CLI lifecycle behavior, not a per-agent component-coverage change.

## 9. Out of scope / do not touch

- **Do NOT fix the `id@marketplace` ref-acceptance bug for `upgrade`/`enable`/`disable`/`remove`** (raw &#34;file not found&#34; instead of &#34;not installed&#34; — `plugin.go:162-318`). That is tracked separately as **#168**; leave `splitPluginRef`-on-`args[0]` for that issue. Do not pre-empt or partially implement it here.
- Do not change `pluginUpgradeRun`&#39;s field-preservation logic except as incidentally required — `upgrade` already read-modify-writes the existing TOML (`plugin.go:170,205-211`) and preserves `Agents`/`Disabled`. (If anything, it&#39;s the correct model; mirror its read-modify-write shape in install rather than rewriting upgrade.)
- Do not introduce any new dest→source write path or call `source.Write*` directly — these are direct canonical-source edits, and the existing `readPluginTOML` + `iox.AtomicWrite` pattern is the right tool.
- Do not touch `walkSecretFields`, adapter `Render`/`DestWriter`, the marketplace fetch/projection code, or `source.PluginSpec`&#39;s shape.

## 10. References

- **Review report:** `agentsync-v1.0-review-FINAL.md` — §B item 5 (&#34;`plugin install` / `disable→enable` silently widen or reset a plugin&#39;s `agents` allowlist&#34;), §D.2 CLI &#34;High&#34; (`plugin` allowlist widening, B#5). Finding refs: `B#5`, `D2-plugin-allowlist-widen`.
- **Code:**
  - `internal/cli/plugin.go:287-288` — disable empties `Agents`
  - `internal/cli/plugin.go:249-252` — enable re-broadens to `[&#34;*&#34;]`
  - `internal/cli/plugin.go:95-148` — `installPluginInto` hard-resets `Agents`/`Update`/`Disabled`
  - `internal/cli/plugin.go:42-52` — `pluginTOMLSpec` shape
  - `internal/cli/plugin.go:319-324` — the existing &#34;no cheerful no-op&#34; convention to mirror
  - `internal/source/schema.go:158-173` — canonical `PluginSpec`
  - `internal/cli/plugin_test.go` — existing plugin tests + `makeLocalMarketplace`/`runCLI`/`mustRun` helpers
- **Docs:** `website/src/content/docs/guides/plugins.mdx:69-93`, `CHANGELOG.md`, `website/src/content/docs/reference/cli.mdx`, `docs/user-guide.md`.
- **Invariants:** `CLAUDE.md` — &#34;Models must stay faithful to their on-disk artifacts,&#34; &#34;Secret-handling invariants,&#34; &#34;Keep the docs in sync.&#34;

---
📋 _Part of the v1.0 remediation epic #178 — see it for the full wave plan and dependency graph._

spxrogers:
## Summary

Wave 2 of the **v1.0 remediation epic #178** — the carried-over high-severity
correctness bugs plus the new-in-v1.0 git-backup/revert data-loss & safety fixes.
14 sub-issues, one commit each, each with artifact-anchored tests and same-commit
docs. Based on `main` at the Wave-1 merge (`d9615b7`); this PR is the 14 Wave-2
commits only.

**Closes #140, #142, #144, #148, #125, #127, #130, #134, #136, #139, #143, #146, #149, #126.**

### What each issue does + how its acceptance criteria were met

**Adapters — silent-drop / round-trip fidelity (C1):**
- **#144 fix(codex)** — Ingest re-populates `fm["name"]` from the TOML `name` unconditionally on presence (divergent name survives round-trip); `renderSubagents` errors on two distinct stems resolving to the same effective name. Tests: `TestIngest_RoundTripsSubagent_DivergentName`, `…_RepopulatesName_WhenMatchingFilename`, `TestRender_Subagent_DuplicateEffectiveName_Detected`. (Code half of Wave-3 doc cell #150.)
- **#148 fix(opencode)** — Ingest stops deleting `mode` and only captures agentsync-owned agents/commands (ownership via the apply state file); render preserves an ingested `mode`, defaults to `subagent` only when absent. Spec-complete on-disk fixtures: `TestIngest_PreservesAgentMode`, `TestIngest_OwnershipFilter_SkipsUnmanagedAgentsAndCommands`, `TestRoundTrip_OpenCode_SpecComplete`.
- **#125 fix(opencode)** — subagent/command render passes legal OpenCode frontmatter keys through verbatim and emits a `SkipReduced` for every unrepresentable key (no key is both absent-from-output and absent-from-skips). On-disk fixtures assert the invariant list-driven.
- **#127 fix(continue)** — command Ingest captures identity from the prompt block's frontmatter `name` (filename fallback), fixing the silent rename of foreign blocks. Artifact-anchored `TestIngest_Command_NameFromFrontmatter` (re-render lands `bar.md` on disk) + `…_FilenameFallback`.
- **#134 fix(gemini)** — subagent render switched to a drop-list (`{tools,color}`); native keys (`kind/temperature/max_turns/timeout_mins/mcpServers`) pass through; the Skip lists only genuinely-dropped keys. On-disk `TestRoundTrip_Subagent_NativeKeysSurvive`.
- **#136 fix(windsurf)** — ingest strips any leading `trigger:` fence (not just `always_on`) so re-apply can't double-fence; foreign-trigger warns, fence-less does not; workflow bodies also de-fenced. Artifact-anchored `TestRoundTrip_ProjectRule_NonAlwaysOnTrigger_NoDoubleFence` + unit table.
- **#130 test(roo)** — new `apply_test.go` certifies the documented `.roo/mcp.json` merge claims: foreign-server preservation, `OwnedKeys` orphan removal, and `>2^53` `json.Number` fidelity (asserted via `UseNumber()` so a rounding regression can't hide). Test-only.

**CLI / secrets write-path data-loss:**
- **#140 fix(cli)** — `plugin disable`/`enable` no longer erase/re-broaden the `agents` allowlist; re-`install` read-modify-writes to preserve `Agents`/`Update`/`Disabled` (first install stays byte-identical), and surfaces what it kept. On-disk `plugins/<id>.toml` assertions.
- **#142 fix(secrets)** — `secrets set a.b=v` refuses to clobber an existing scalar (and a leaf overwriting a table), threaded before encryption so `secrets.age` is byte-for-byte unchanged on refusal; the `set` path now runs `ValidateVaultTOML`. Error text never contains secret bytes. End-to-end vault tests.
- **#139 fix(cli)** — `agentsync version` delegates to cobra's precompiled FuncMap version renderer (re-dispatches `--version` on root) instead of hand-parsing `text/template`, so the two are byte-identical by construction. Raw-byte parity + NoArgs + help tests. Verified live: `version` == `--version`.

**Git-backup / revert (new-in-v1.0 data-loss & safety), built on Wave-1 #128:**
- **#146 fix(git)** — `Restore` folds `SnapshotDirtyTracked` internally (safe-by-construction for every caller), removes the redundant CLI snapshot, and emits a recovery hint (orig + snapshot hash) on partial-reset failure. Engine-level on-disk tests.
- **#149 fix(git)** — `HasNestedRepoBelow` is symlink-aware and re-checked before `Restore`, on the commit path (warn), and in `doctor` (warn), so a foreign repo cloned under an owned root after init is never hard-reset. Real nested-repo fixtures asserting the foreign file is byte-unchanged (default/strict/dry-run).
- **#143 fix(git)** — a pre-apply baseline checkpoint makes the **first** apply revertible (≥2 commits, apply's parent is the baseline); baseline stages the dir's full pre-existing contents; best-effort with a loud warning on failure. Artifact-anchored `TestApply_GitBackupBaselineRevertsFirstApply` + siblings.
- **#126 fix(git)** — `.git` perms re-tightened (best-effort, warn) on every owned commit path, not just at init; init made atomic-ish (owned repo always has the NOTICE); `mode=on` new-dir cleartext notice; revert-snapshot cleartext caution; inert project-scope git-backup override dropped. `TestNoPushSurface` unchanged.

## Type of change

- [x] Bug fix
- [x] Tests / CI / tooling  (#130 is test-only; every fix adds tests)
- [x] Docs  (same-commit CHANGELOG / capability-matrix / user-guide / architecture / cli.mdx / plugins.mdx per issue)

## Test plan

`just`/podman aren't available in this environment, so the hermetic recipes were
run via their exact underlying commands with the container guard satisfied
(`AGENTSYNC_TEST_IN_CONTAINER=1`, `GOTOOLCHAIN=go1.26.2`):

- **Full race suite** (`just test` equivalent): `go test -race -count=1 ./...` → **green** (28 pkgs ok, no failures, no data races).
- **Lint** (`just lint`): `gofmt -s` + `gofumpt` + `go mod tidy` produced no changes; `golangci-lint@v2.12.2 run ./...` → **0 issues**.
- **Build**: `go build ./...` clean.
- **Live drive**: built the binary; `agentsync version` and `agentsync --version` are byte-identical; `version foo` is rejected; command tree intact.
- Each of the 14 issues ships artifact-anchored tests that seed a spec-complete on-disk fixture and assert the on-disk result (see per-issue notes above).

- [ ] `just test-release` is green (the release bar) — could not run the full hermetic **container** gate here (no podman/docker-in-docker); ran the equivalent `-race` suite + lint directly. CI's `test-release` should confirm.
- [x] `just lint` is clean

## Checklist

- [x] Conventional commit messages with a scope.
- [x] Tests added/updated for the behavior changed (artifact-anchored per CLAUDE.md).
- [x] `internal/secrets` (#142) touched only the CLI vault-write path — no `capture.Capture` / `walkSecretFields` / `source.Write*` / `secrets.Resolved` change; invariants re-read and not weakened. #126 keeps the local-git cleartext-history exception bounded by the never-push guard (`TestNoPushSurface` unchanged).
- [x] Docs updated in the same commit as each behavior/CLI/coverage change.

### Reviewer notes / deviations to scrutinize
- **#148 layering:** the ingest ownership filter reads the apply state file (`targets.json`) from inside the opencode adapter (adds `internal/state`+`internal/paths` imports to the adapter). Defensible per the issue's "use the same ownership source as the pipeline," but it's an adapter→state dependency worth a look; noted as a class-wide gap (other adapters don't filter ingest ownership).
- **#146 API:** `git.Repo.Restore` signature changed to `(revertHash, snapshotHash string, err error)`; all in-repo callers updated.
- **Commits show as *Unverified*:** the environment's SSH signing key file is empty, so no signature could be produced (identity is correct: `Claude <noreply@anthropic.com>`).

---
_Generated by [Claude Code](https://claude.ai/code/session_011SS7xSq6VdbA2QKEDymcgN)_


spxrogers:
## `/review-loop` — converged in 3 rounds ✅

Ran the repo's four-lens adversarial review loop (correctness · adversarial · API design · test rigor) against this PR. Converged to **all four `CLEAN — ship it`** after fixing every actionable finding. Three review commits were added on top of the 14 issue commits:

**Round 1** (`92b1694`) — findings fixed:
- **`fix(git)`**: the #143 pre-apply baseline used `git add -A`, so a first apply into `~/.claude` would durably commit `.credentials.json` + `projects/**` transcripts into local git history. Narrowed to stage only the NOTICE + the *managed* files' pre-apply content; #128 already preserves untracked siblings on revert, so this loses no recoverability (#143's explicitly-permitted "out-of-contract" alternative). Removed the now-unused `StageAll`.
- **`docs(architecture)`+`fix(opencode)`**: documented the `opencode ingest → internal/state` layering exception in §10 (deliberate, opencode-scoped, test-guarded); warn instead of silently capturing nothing when the apply-state file is absent.
- **`fix(git)`**: corrected the #149 nested-repo guard rationale (the new `Restore` is a surgical, untracked-safe delta apply — not a hard reset) across the revert/commit warnings + the `applyRestoreDelta` comment.
- **`fix(cli)`**: a re-install over a *corrupt* `plugins/<id>.toml` now refuses instead of silently re-broadening the `agents` allowlist.
- Doc NITs: `version.go` SetArgs precondition; `checkpoint` error-return rationale.

**Round 2** (`6f149c4`) + **Round 3** (`e3a5886`) — closed test-coverage gaps: the data-loss-on-rollback guard for the baseline's managed-capture branch (break-verified), the corrupt/empty/populated `plugins/<id>.toml` boundary (three-sided), and the opencode absent-state warning; plus a vacuous-pass hardening.

**Final local gate:** `go test -race ./...` green (28 pkgs, no races), `golangci-lint` 0 issues, `gofmt`/`gofumpt`/`go mod tidy` clean.

_(Note: commits show as GitHub-"Unverified" — the environment's SSH signing key file is empty, so no signature could be produced; the committer identity is correct.)_

---

Issue #176, opened 2026-06-30
fix(core): carried-over core/git nits — untrusted Event, Apply dispatch DRY, sort, doc comments, shortHash dedup


spxrogers:
## Metadata

- **Severity:** nit (one bundled finding, `D1-git-detect-state-with-nonnil-error`, is severity *low* — see Problem §6)
- **Area:** core (spans `internal/source`, `internal/render`, `internal/adapter`, `internal/iox`, `internal/git`, `internal/cli`)
- **Wave:** 4
- **Themes:** C2 (doc/code consistency); DRY cleanup
- **Depends on:** none
- **Can run in parallel with:** everything — this is an independent low-risk nit batch. **Coordinate the `internal/git/git.go` and `internal/git/restore.go` regions with the Wave-2 git-safety issues #146 / #149 (and the substantive set #128/#143/#126) to avoid churn** — if those are in flight, rebase the small edits here on top of them.
- **Labels:** `severity:nit`, `area:core`, `wave:4`

## Summary

A grab-bag of carried-over core/git nits accumulated across the apply/git subsystems: an untrusted-origin field not modeled as `untrusted.Text`, ten copies of an identical adapter `Apply()` dispatch, a hand-rolled bubble sort mislabeled &#34;insertion sort&#34;, two doc comments that misstate the code they describe, three identical 7-char hash abbreviators, a `git.Detect` that returns a meaningful `State` alongside a non-nil error, a non-exhaustive rename-handling switch, an unbounded per-apply git-status cost, and a sticky-mode prompt that persists `mode=on` after a skipped init without a note. None are correctness bugs today; each is either a clarity/consistency defect or a fail-safe / cost hardening. &#34;Done&#34; = each item below is reconciled so code matches its doc, duplication collapses to one helper, and the two fail-safe items either change behavior or gain an explicit contract comment — with the docs-sync contract honored.

## Background &amp; context (for an agent with zero context)

`agentsync` is a single-machine Go CLI that renders a canonical config (`~/.agentsync/`, project scope `.agentsync/`) into each AI agent&#39;s native config and captures native edits back. Relevant subsystems for this issue:

- **`internal/source`** — the canonical model + loaders. `loadHooks` reads `hooks/.toml` files; the file *name* (an on-disk, user/marketplace-controllable origin) becomes `Hook.Event`.
[...]

Issue #175, opened 2026-06-30
fix(adapters): normalize roo/cline/generic MCP transport, fix roo/noop Detect probes, narrow IDE-dir version/walk lows, and add a per-row transport-normalization note to the capability matrix


spxrogers:
## 1. Metadata

- **Severity:** low (bundle of low + nit findings; the C5 matrix-harmonization deliverable is the medium-leaning net-new piece)
- **Area:** adapters
- **Wave:** 4
- **Themes:** C5 (MCP transport-normalization harmonization), with C1 ownership-asymmetry and Detect/Capabilities (core #6) touchpoints
- **Depends on:** none (not blocked)
- **Can run in parallel with:**
  - #172 (cursor adapter) — also edits capability-matrix transport rows; coordinate the matrix edit
  - #166 (gemini `$` re-expansion) — also a C5 matrix-row touchpoint
  - #154 (VersionedDirs guard tightening) — also edits `internal/adapter/generic/homedir.go`; coordinate that file
- **Labels:** `severity:low`, `area:adapters`, `theme:C5`, `wave:4`

## 2. Summary

A cluster of adapter low/nit cleanups plus the cross-cutting **C5** deliverable. Several adapters silently or only-half-document MCP transport normalization on round-trip (roo flips an untyped-url server `stdio→remote`; cline and the generic breadth tier collapse `sse→http`), `Detect()` probes are wrong or dead in `roo` and `noop`, the `noop` package is under-tested and unguarded, claude&#39;s MCP/LSP `Extra` ingest loses precision on integers `&gt;2^53`, and the whole-IDE-dir VersionedDirs design tradeoff is undocumented along with a potentially-unbounded init-time tree walk. &#34;Done&#34; means: transport handling is consistent and **backed by round-trip tests** across roo/cline/generic; the capability matrix carries a coherent **per-row &#34;transport normalization&#34; note** for every MCP-bearing adapter (closing the acknowledged-vs-silent gap in one pass); the roo and noop `Detect`/name/doc/test gaps are fixed; claude `Extra` ingest preserves large integers; and the IDE-dir tradeoffs and walk cost are documented (and the walk optionally bounded).

## 3. Background &amp; context (for an agent with zero context)

`agentsync` is a single-machine Go CLI that renders one canonical config (`~/.agentsync/`, the project-scope `.agentsync/` tree here) into each agent&#39;s native config and captures native edits back. Read `/home/user/agentsync/CLAUDE.md` first.

[...]

Issue #173, opened 2026-06-30
qa(release): empirically verify build reproducibility by diffing two real releases


spxrogers:
## Metadata

- **Severity:** low
- **Area:** release
- **Wave:** 4
- **Themes:** C2 (doc/code drift on contract surfaces — the `.goreleaser.yaml` &#34;reproducible, identical&#34; comment is an unverified capability claim); release-engineering QA / &#34;doc claims are not self-certifying&#34; (CLAUDE.md §&#34;Models must stay faithful to their on-disk artifacts&#34; rule 3).
- **Depends on (blocked-by):** **#138** — *chore(release): pin goreleaser version, restore archive mtime pinning, add checksum signing, nfpms URL guard*. #138 decides whether archive mtime pinning is restored or the reproducibility comment is softened; this issue empirically validates whichever outcome #138 chose.
- **Can run in parallel with:** Nothing in this wave — this is the QA verification counterpart to #138 and must run *after* #138 lands so it validates the chosen outcome (restored mtime pinning vs. corrected comment), not the pre-#138 state.
- **Labels:** `severity:low`, `area:release`, `wave:4`

## Summary

The agentsync release config (`.goreleaser.yaml`) claims its builds and archives are reproducible (&#34;a given tag builds bit-for-bit identically&#34;, &#34;the (reproducible, identical) assets&#34;). That claim was only ever assessed by *reading* the config — never by actually running two releases and diffing the produced archives and `checksums.txt`. A v1.0 review further noted that the archive-mtime regression attributed to PR #115 was *inferred from the PR diff, not empirically reproduced*. &#34;Done&#34; means: a concrete, repeatable procedure (run twice, diff `dist/`) has been executed, its result recorded, and the outcome reconciled with #138 — either the artifacts diff clean (claim verified, comment kept/tightened) or they differ (claim false, and the `.goreleaser.yaml` comment is corrected to match reality, since #138 owns the config fix itself).

## Background &amp; context (for an agent with zero context)

**What agentsync is.** agentsync is a single-machine Go CLI (`cmd/agentsync`) that centrally manages AI coding-agent configurations. It ships to end users through several package managers; the *entire* release is produced by one [goreleaser](https://goreleaser.com) v2 run.

**Where release engineering lives.**

[...]

Issue #172, opened 2026-06-30
fix(continue+cursor): round-trip MCP version/schema, disambiguate command+url, and close Cursor adapter test/skip gaps


spxrogers:
## Metadata

- **Severity:** low (mix of `low` and `nit` findings)
- **Area:** adapters (`internal/adapter/continuedev`, `internal/adapter/cursor`)
- **Wave:** 4
- **Themes:** C5 (MCP transport classification / normalization), C1 (round-trip fidelity)
- **Depends on:** none (independent low backlog)
- **Can run in parallel with:** any other low-severity adapter issue. Touches only `internal/adapter/{continuedev,cursor}/*`. The Cursor remote-`type` reject-vs-ignore verdict is *also* probed by QA issue **#164** — coordinate the verdict, but the code/test changes here do not block on it.
- **Labels:** `severity:low`, `area:adapters`, `theme:C5`, `wave:4`

---

## Summary

A cluster of low-severity correctness and test-coverage gaps in the **Continue** and **Cursor** MCP adapters. Continue regenerates the per-server block `version`/`schema` header instead of round-tripping it, and silently classifies a server carrying *both* `command` and `url` as stdio (dropping the URL). Cursor emits a `type` key on *remote* MCP servers that its documented remote schema (`url`+`headers`, no `type`) does not define, and several of its adapter behaviors (Extra passthrough across a full Render→Apply→Ingest cycle, orphan-cleanup of the `version` key, `is_background` rendering) are untested or covered only by false-passable substring assertions. &#34;Done&#34; means: Continue round-trips `version`/`schema` and refuses to silently drop a URL; Cursor either stops emitting the off-spec remote `type` (or documents it pending the #164 real-Cursor verdict); and the named tests exist and assert precisely.

---

## Background &amp; context (for an agent with zero context)

`agentsync` is a single-machine Go CLI that renders one canonical config (`source.Canonical`, TOML+markdown under `~/.agentsync/`) into each AI agent&#39;s native config, and captures native edits back. Each agent has an adapter implementing the `Adapter` interface: `Render(secrets.Resolved, scope, project) ([]FileOp, []Skip, error)` projects canonical → native FileOps; `Apply(ops, DestWriter)` writes them; `Ingest(scope, project) (source.Canonical, error)` reads native files back into canonical. A **round-trip** is Render→Apply→Ingest (or Ingest→Render); fidelity means nothing is silently lost.

The two adapters in scope:

[...]

PR #196, opened 2026-07-20
epic #178 — Wave 4: cross-cutting hardening, adapter/core/CLI lows, artifact-anchored fidelity & QA verification

### LINKED ISSUE #153: test(C1): convert model-seeded round-trips and safety-promise tests to artifact-anchored fidelity tests
> [!IMPORTANT]
> **Update since this issue was filed — PRs #73 and #180 merged after the epic (#178) was created.**
> **PR #73** ("fix: skip unsupported Claude LSP settings", which resolved the pre-existing #66) **removed Claude's LSP support entirely**: it deleted `internal/adapter/claude/lsp.go`, dropped `CapLSP` from Claude's `Capabilities()` bitmask, stopped writing/reading `settings.json#/lspServers`, and now emits an `adapter.Skip` for Claude LSP — with `capability-matrix.md`, `architecture.md`, `components.md`, `user-guide.md` and the tests updated in the same commit. **PR #180** re-homed the render-pipeline sibling-section / orphan-cleanup regression tests onto Codex and touched `internal/render/pipeline.go`.
>
> Because these commits changed `claude/{render,ingest,claude,paths,render_test,ingest_test}.go`, several `docs/*.md`, and `render/pipeline.go`, **the exact `file:line` references below have drifted — locate code by symbol, not by line number.**
>
> **Impact on THIS issue:**
> - The Claude slice (§4a) targeted `TestIngest_RoundTripsHooksButIgnoresSettingsLSP` with a fixture containing an **LSP server**. Claude LSP is gone and #73 already reworked `ingest_test.go`, so make this a **hooks-only** artifact-anchored test: an on-disk `settings.json` with an unmodeled hook key + a non-`command` handler; **drop the LSP dimension**. Confirm the current test name (the `…HooksAndLSP` test was renamed/gutted by #73). This still backs #124.

---

## 1. Metadata

- **Severity:** high
- **Area:** tests
- **Wave:** 4
- **Themes:** C1 (the impoverished-subset / silent-drop class)
- **Depends on (blocked-by):**
  - #124 — `fix(claude): preserve unmodeled hook fields and stop corrupting non-command handlers on import→apply`
  - #128 — `fix(git): stop agentsync revert from deleting the user's untracked files in Restore`
  - #144 — `fix(codex): re-populate subagent frontmatter name on ingest and detect colliding effective names`
  - #148 — `fix(opencode): add ingest ownership filter and preserve primary/all agent mode across round-trip`
  - #125 — `fix(opencode): pass through or Skip unmodeled subagent/command frontmatter keys`
  - #127 — `fix(continue): ingest command name from frontmatter, not the filename`
  - #134 — `fix(gemini): pass through captured native subagent frontmatter instead of stripping it on apply`
  - #136 — `fix(windsurf): strip non-always_on trigger frontmatter on ingest so re-apply doesn't produce a malformed double-fence rule`
  - #143 — `fix(git): take a pre-apply baseline checkpoint so the first apply is revertible`
- **Can run in parallel with:** Once the dependency fixes have landed (Waves 1–2), this umbrella can be **sharded across many agents by area** — each adapter package, `internal/git`, and `internal/cli` are independent shards. The per-fix fidelity test for a given fix must land *after* that fix.
- **Labels:** `severity:high`, `area:tests`, `theme:C1`, `wave:4`

---

## 2. Summary

Across the codebase, bidirectional round-trip tests and destructive-safety tests are seeded from the **parsed canonical model or the test's own in-memory fixtures**, not from spec-complete **on-disk artifacts**. Because the model models a subset of its artifact and every downstream piece faithfully mirrors that subset, a round-trip over an impoverished model is trivially "lossless" and the silent-drop / silent-deletion class of bug passes CI undetected. "Done" means the carried-over adapter/CLI round-trips **and** the new v1.1 git-feature safety tests are converted to artifact-anchored fidelity tests that start from a spec-complete on-disk fixture and assert the **on-disk result survives byte-for-byte** (or, for destructive paths, that the file/mode/untracked-file actually survives or is actually deleted). This is the single highest-leverage hardening theme; treat it as an **exhaustive sweep**, not just the enumerated instances.

---

## 3. Background & context (for an agent with zero context)

`agentsync` is a single-machine Go CLI that renders a canonical config (`~/.agentsync/`, modeled by `source.Canonical` in `internal/source`) into each agent's native files (`apply`), and captures native edits back (`import` / `reconcile`). v1.1 added **git-backed destination directories**: each managed user-scope root is a git repo, `apply` checkpoints what it wrote, and `agentsync revert` rolls a root back to a prior checkpoint (`internal/git`, driven from `internal/cli/gitbackup.go` and `internal/cli/revert.go`).

This issue is **test work**. It does not change product behavior; it hardens the test suite so the fidelity and safety promises the other Wave-1/2 fixes establish are actually *proven* against the on-disk artifact rather than against an oracle that can't see the gap.

### The governing invariant (CLAUDE.md, verbatim)

From the **"Models must stay faithful to their on-disk artifacts — non-negotiable"** section:

> A whole class of silent bug: a canonical struct models a *subset* of the artifact it claims to represent, and every downstream piece (loader, adapters, writer, projection) faithfully mirrors that subset — so nothing fails a test, no invariant breaks, no lint fires, yet data is dropped. … It survived because the round-trip tests' oracle was the parsed model, not the filesystem — and a round-trip over an impoverished model is trivially "lossless." **The on-disk artifact — what the upstream spec actually defines — is the source of truth, never the struct that parses it.**

And the first of the three rules that follow:

> **Fidelity tests anchor to the artifact, not the model.** For any component backed by a file tree or a structured file, the round-trip test MUST start from a *spec-complete* on-disk fixture (a skill with bundled files; an MCP server with an unmodeled native key; memory with `fragments/`) and assert the *on-disk result* survives — byte-for-byte where applicable. Asserting "the parsed model round-trips" proves nothing about what the model can't see.

This issue *is* the enforcement of that rule across the whole bidirectional + destructive surface.

### Test conventions (CLAUDE.md, "Code conventions")

- **Stdlib testing only** — no testify/gomega. Table-driven with a `name` field and `t.Run`. `t.Helper()` in helpers that call `t.Fatal/Errorf`.
- **Filesystem in tests** is `afero.NewMemMapFs()` or `t.TempDir()`, never `os.UserHomeDir()`. FS-touching tests must run in the container: call `testenv.RequireContainer(t)` (the `internal/git` tests already do — see `TestRestoreAppendOnly`).
- **Errors** wrap with `fmt.Errorf("doing X: %w", err)`; match with `errors.Is/As`.

### Which packages this touches

- Adapters: `internal/adapter/{claude,roo}` (plus the carried-over set: `opencode`, `codex`, `gemini`, `cursor`, `windsurf`, `cline`, `generic`).
- `internal/git` (`commit.go`, `restore.go`, `log.go`, `git_test.go`).
- `internal/cli` (`gitbackup.go`, `revert.go`, `import_test.go`, `revert_test.go`, `mcp.go`).

---

## 4. Problem

The defect is a **test-oracle gap**, not (in this issue) a product bug. Each instance below is grounded in the current code.

### 4a. Claude LSP + hook ingest tests are model-anchored (`internal/adapter/claude/ingest_test.go:229`)

`TestIngest_RoundTripsHooksButIgnoresSettingsLSP` builds the input as an in-memory `source.Canonical`, renders it, applies via `PassThroughWriter{}`, ingests, and asserts only that the *parsed model* round-trips:

```go
func TestIngest_RoundTripsHooksButIgnoresSettingsLSP(t *testing.T) {
	tmp := t.TempDir()
	in := source.Canonical{
		Hooks: []source.Hook{{Event: "PreToolUse", Matcher: "Bash", Type: "command", Command: "echo before"}},
		LSPServers: []source.LSPServer{{ID: "gopls", Spec: source.LSPServerSpec{Command: "gopls", Args: []string{"serve"}}}},
	}
	...
	if len(out.Hooks) != 1 || out.Hooks[0].Event != "PreToolUse" { ... }
	if len(out.LSPServers) != 1 || out.LSPServers[0].ID != "gopls" { ... }
}
```

The fixture is *exactly* what the model can represent, so it can never observe an unmodeled hook field being dropped or a non-command handler being corrupted — this is the missing test that would have caught #124.

### 4b. Skill import is SKILL.md-only — bundled files unverified (`internal/cli/import_test.go:670`)

`TestImport_SkillFromClaude` seeds a skill dir with only `SKILL.md` and asserts only that the body string is captured:

```go
skillDir := filepath.Join(tmp, ".claude", "skills", "deploy")
os.WriteFile(filepath.Join(skillDir, "SKILL.md"), []byte("---\nname: deploy\n...\n---\nSteps to deploy.\n"), 0o644)
...
data, _ := os.ReadFile(filepath.Join(tmp, ".agentsync", "skills", "deploy", "SKILL.md"))
if !strings.Contains(string(data), "Steps to deploy") { ... }
```

`scripts/`, `references/`, `assets/` written to the dotfiles repo on import are never asserted to round-trip — exactly the `source.Skill` impoverished-subset failure that CLAUDE.md calls out by name.

### 4c. Roo command round-trip starts from the parsed model (`internal/adapter/roo/ingest_test.go:20`)

`TestRoundTrip_ProjectScope` builds the command in memory and checks ingested *model* fields:

```go
Commands: []source.Command{{Name: "deploy", Frontmatter: map[string]any{"description": "Deploy", "argument-hint": "<env>", "allowed-tools": "Bash"}, Body: "Deploy it.\n"}},
...
if cmd.Name != "deploy" || cmd.Body != "Deploy it.\n" { ... }
```

Disk-side ingest IS covered elsewhere; what is missing is a **byte-stability oracle** — render → read the on-disk command file → ingest → re-render → assert the on-disk bytes are identical.

### 4d. `mcp add` tests never read back `--env` / `--agents` (`internal/cli/mcp.go`, mcp add tests)

The `mcp add` tests set `--env`/`--agents` but never read those values back out of the canonical model, so their round-trip into `source.Canonical` is unverified.

### 4e. The new git-feature safety tests are model-seeded / impoverished-oracle

- **`TestRestoreAppendOnly` (`internal/git/git_test.go:517`)** verifies content restore + deletion, but **every committed fixture is a `commitFile` 0o600 regular text file**. It seeds with:

  ```go
  commitFile(t, r, dir, "a.txt", "v1", "apply 1")
  commitFile(t, r, dir, "a.txt", "v2", "apply 2")
  commitFile(t, r, dir, "b.txt", "added", "apply 3")
  ```

  Its oracle cannot observe (i) an **untracked** file being deleted by the hard reset (#128), nor (ii) an **executable / mode bit** being dropped, because no fixture file is untracked or executable. `Restore` runs a `HardReset` then `SoftReset` (`internal/git/restore.go:90-95`); a hard reset is precisely what can blow away untracked files and is where mode fidelity must be proven.

- **Stage is content-diff-driven, not write-driven (`internal/cli/gitbackup.go:86`, `internal/git/commit.go:20`).** `Stage` calls `wt.Add(rel)` per file; go-git's `doAddFile` early-returns when the worktree status equals the index. Combined with `CommitStaged` returning `("", nil)` on a clean index (`commit.go:62-72`), the captured tree is the content diff, not "every file the apply wrote." No test asserts that **every member of `written` under a root appears in the resulting checkpoint tree** (they coincide today but diverge on an identical-content rewrite or a mode-only change).

- **`init`-over-non-empty-dir** and the **apply→hand-edit-untouched-tracked-file→revert seam** are untested. The checkpoint stages only this-run rels + the notice + tracked deletions (`gitbackup.go:83-95`); the only net for a hand-edited *untouched* tracked file is `revert`'s later `SnapshotDirtyTracked` before the hard reset (`revert.go:228`). That cross-component invariant is load-bearing and unproven end-to-end.

- **Shared-dir multi-owner revert + folded-child revert** are unproven end-to-end. `revertRoot` warns when a root is shared (`revert.go:247-250`) and redirects a folded child to its parent (`revert.go:186-191`), but no `apply`+`revert` test exercises a shared root asserting both owners' files roll back together and the warning fires, nor a folded child root.

- **`revert` global-lock serialization + dry-run lock-skip are untested (`internal/cli/revert.go:71-74`).** `apply` has lock tests; `revert` acquires the *same* lock for mutating runs and deliberately skips it for `--dry-run`:

  ```go
  if dryRun {
  	return run()
  }
  return withGlobalLock(home, run)
  ```

  `revert` is data-loss-capable (hard reset + index writes); a refactor dropping `withGlobalLock`, or accidentally locking the dry-run path, would pass CI. Separately, the dry-run path reads go-git index/HEAD (`previewRevert` → `repo.Plan`/`repo.IsClean`, `revert.go:256-266`) **without** the lock, so it can read a torn index while a concurrent apply rewrites `.git/index` — an unsound "read-only ⇒ no lock" assumption (track as a documented limitation if not fixed under the dep).

- **`--all` + `--to` rejection is untested (`internal/cli/revert.go:48`).** The guard exists:

  ```go
  if all && toRef != "" {
  	return fmt.Errorf("--to names a checkpoint in one repo and can't apply across --all; ...")
  }
  ```

  but `TestRevert_Errors` covers a *different* branch; this destructive-command arg guard could be dropped by a refactor undetected.

- **`--to <ref>` accepts any resolvable commit with no ancestor/checkpoint validation (`internal/cli/revert.go:222`, `internal/git/log.go:58`).** `repo.Resolve` is go-git `ResolveRevision`, which returns *any* commit that parses; `Restore` then force-sets the worktree to that tree. A fat-fingered hash or a sibling ref silently restructures live config. (The *code* fix may be a separate issue; this issue must add the **test** that a non-ancestor/non-checkpoint `--to` ref is rejected.)

- **Resolved cleartext actually lands in destination history is unverified.** The v1.1 rationale that resolved secrets are acceptable in the *local-only* git history is asserted nowhere; no test proves a `${secret:…}`-resolved value is present in the destination checkpoint (so that, conversely, the *canonical* leak guards remain the only thing standing between cleartext and the dotfiles repo).

### 4f. The assurance-gap caveat (report §F)

Because round-trips remain model-seeded and restore-safety promises are pinned by impoverished oracles, **the absence of a failing test does not certify fidelity or safety** for any bidirectional or destructive path. The instances above were found by reading, not by an exhaustive sweep. The C1 work is not "done" until every `import`/`reconcile` path and every `revert`/`Restore` path is artifact-anchored.

---

## 5. Required changes

Do these as an **ordered sweep**, sharded by area. Each conversion replaces a model/self-fixture oracle with a **spec-complete on-disk fixture** and asserts the on-disk result survives (byte-for-byte where the artifact is text/structured).

1. **Adopt a shared fidelity helper** (per area, or one in a small test-support file): given a destination dir, render → capture the on-disk artifact bytes → ingest → re-render → assert the destination bytes are **byte-identical** for text/structured files; for file trees (skills, memory `fragments/`), walk the tree and assert the full set of relative paths + bytes survive. Do NOT compare parsed models as the oracle.

2. **Claude (`internal/adapter/claude/ingest_test.go`)** — convert `TestIngest_RoundTripsHooksButIgnoresSettingsLSP` to start from a **spec-complete on-disk `settings.json`** that contains: a `command` hook **with an unmodeled sibling key**, a **non-command** hook handler, and an LSP server with an **unmodeled native key** (carried in `Extra`). Assert the on-disk `settings.json` survives the import→re-apply round-trip byte-for-byte (this backs #124).

3. **Skill import (`internal/cli/import_test.go`)** — add an import test seeded from a **spec-complete on-disk skill**: `SKILL.md` plus `scripts/deploy.sh` (executable), `references/notes.md`, `assets/logo.png`, and a nested file. Assert every bundled file round-trips into `.agentsync/skills/deploy/…` with identical bytes (and that the executable bit survives).

4. **Roo (`internal/adapter/roo/ingest_test.go`)** — add a **byte-stability oracle** for the command: render → read the on-disk command file bytes → ingest → re-render → assert the bytes are identical.

5. **`mcp add` (`internal/cli/mcp.go` tests)** — extend the `mcp add` tests to **read back** `--env` and `--agents` from the canonical model and assert they round-trip.

6. **`internal/git` — Restore fidelity:**
   - Add an **untracked-survival** test: commit fixtures, drop an **untracked** file into the worktree, `Restore` to a prior checkpoint, assert the untracked file is **still present** with identical bytes (backs #128).
   - Add a **mode-bit fidelity** test: commit a fixture file at one mode, commit again with the **executable bit set** (e.g. a hook script that became `+x`), `Restore` across the change, assert `os.Stat(...).Mode().Perm()` matches the target checkpoint.

7. **`internal/git`/`internal/cli` — write-driven capture:** add a git-anchored test asserting that **every member of the `written` set under a root appears in the resulting checkpoint tree**, including an **identical-content rewrite** and a **mode-only change** case (proves capture is write-driven, not content-diff-driven; backs `gitbackup.go:86` / `commit.go:20`).

8. **`internal/cli` — apply/revert end-to-end seams:**
   - `init`-over-non-empty-dir: run apply into a dir that already has unmanaged files, assert the repo inits and the first checkpoint is created without clobbering pre-existing files (interplays with #143's pre-apply baseline).
   - apply → hand-edit an **untouched tracked** file → `revert` → assert the edit **survives as a snapshot commit** (the `SnapshotDirtyTracked` net at `revert.go:228`).
   - **shared-dir multi-owner revert** (e.g. `~/.agents/skills` written by codex + a breadth agent): assert both owners' files roll back together **and** the cross-agent warning at `revert.go:247-250` fires.
   - **folded-child revert**: a child root nested in a parent agentsync repo prints the "revert that parent" notice (`revert.go:186-191`) and is rolled back via the parent.

9. **`internal/cli/revert_test.go` — concurrency + arg guards:**
   - Mirror apply's `TestConcurrentApply_LockSerializes` / `…DryRunDoesNotContend` for `revert`: serialization for mutating runs, lock-skip for `--dry-run` (`revert.go:71-74`).
   - Add a `TestRevert_Errors` subcase for **`--all` + `--to`** (`revert.go:48`).
   - Add a subcase asserting a **non-ancestor / non-checkpoint `--to` ref is rejected** (`revert.go:222`, `git/log.go:58`). If the underlying ancestor-validation code fix is a separate dep, write the test against the post-fix behavior; otherwise mark `t.Skip` with a TODO referencing that issue rather than asserting the unsafe current behavior.

10. **Resolved-cleartext-lands-in-history:** add a test that applies a config whose MCP/hook field references `${secret:K}`, and asserts the **resolved cleartext value is present in the destination git checkpoint** (proving the local-only history rationale is real), while the **canonical source still holds the `${secret:K}` reference** (proving the capture re-reference + leak backstop, per CLAUDE.md secret-handling invariants, were not weakened). Do **not** route any new dest→source write outside `capture.Capture`.

11. **Carried-over adapter sweep** (shard per package; same artifact-anchored pattern): OpenCode (mode + frontmatter passthrough — backs #148/#125); Codex (skill-bundled-files, `developer_instructions` body edge cases, MCP/hooks coexistence, subagent name — backs #144); Gemini (subagent frontmatter passthrough, hook-group, namespaced commands — backs #134); Cursor (`Extra` passthrough); generic-tier `sse`→`http` transport flip; Windsurf (workflow over-capture + `trigger: glob` memory fold — backs #136); Cline (workflow over-capture). For each, the fixture MUST be spec-complete on disk (include the unmodeled native key / bundled file / non-default trigger) and the oracle MUST be the on-disk bytes.

12. **Exhaustiveness:** after the enumerated conversions, do a sweep of the remaining `*_test.go` in every adapter package + `internal/git` + `internal/cli` and convert any remaining `source.Canonical{…}` → render → ingest → compare-model patterns on a bidirectional or destructive path to artifact-anchored form, or leave an explicit `// MODEL-ANCHORED: <reason it's acceptable>` comment where a model oracle is genuinely sufficient (e.g. a pure unit of a helper that never touches disk).

Respect project invariants throughout: dest→source write-backs go only through `capture.Capture`; new secret-bearing fields go only in `walkSecretFields`; renders go through `DestWriter`. (No product code changes are expected from this issue beyond test scaffolding; if a conversion reveals a *new* product bug, file it separately and `t.Skip` with a reference rather than silently asserting buggy behavior.)

---

## 6. Acceptance criteria

- [ ] Every converted round-trip test starts from a **spec-complete on-disk fixture** (not an in-memory `source.Canonical`) and asserts the **on-disk artifact** survives — byte-for-byte for text/structured files, full path+bytes set for file trees.
- [ ] Claude hook+LSP fidelity test starts from an on-disk `settings.json` containing an unmodeled hook key, a non-command handler, and an LSP `Extra` key, and asserts byte-for-byte survival across import→re-apply.
- [ ] Skill import test seeds `SKILL.md` + `scripts/` (executable) + `references/` + `assets/` + a nested file and asserts all bundled files (and the +x bit) round-trip into the dotfiles repo.
- [ ] Roo command test has a byte-stability oracle (render→read→ingest→re-render→identical bytes).
- [ ] `mcp add` tests read back `--env` and `--agents`.
- [ ] `internal/git` has an untracked-file-survives-Restore test and a mode-bit-fidelity Restore test.
- [ ] A git-anchored write-driven-capture test asserts every `written` member appears in the checkpoint tree, including an identical-content rewrite and a mode-only change.
- [ ] End-to-end tests exist for: init-over-non-empty-dir; apply→hand-edit-untouched-tracked→revert (edit survives as snapshot); shared-dir multi-owner revert (both owners roll back + warning fires); folded-child revert (parent notice).
- [ ] `revert` has lock-serialization and dry-run-does-not-contend tests mirroring apply; `--all`+`--to` rejection has a test; non-ancestor `--to` ref rejection has a test (or a `t.Skip` referencing its code-fix issue).
- [ ] A test proves resolved cleartext lands in the destination git history while the canonical source retains `${secret:K}`.
- [ ] The carried-over adapter set (OpenCode, Codex, Gemini, Cursor, generic, Windsurf, Cline) round-trips are artifact-anchored; any remaining model-oracle round-trip on a bidirectional/destructive path is either converted or annotated `// MODEL-ANCHORED: <reason>`.
- [ ] All new/changed tests follow project conventions (stdlib only, table-driven + `t.Run`, `t.Helper()`, `t.TempDir()`/afero, `testenv.RequireContainer` for FS-touching tests).
- [ ] `just test` (and `just test-release` where the container gate applies) passes; `just lint` is clean and any files it rewrote are re-staged.

---

## 7. Tests to add / update

All FS-touching tests call `testenv.RequireContainer(t)`. Use `t.TempDir()` (or `afero.NewMemMapFs()` for pure-FS units). Table-driven with `t.Run` where there are multiple shapes.

- **`TestIngest_RoundTripsHooksButIgnoresSettingsLSP_OnDisk`** (`internal/adapter/claude/ingest_test.go`, replacing the model-anchored body) — fixture: spec-complete `settings.json` with (a) a `command` hook carrying an unmodeled sibling key, (b) a non-`command` handler, (c) an LSP server with an unmodeled native key. Asserts the on-disk `settings.json` is byte-identical after import→re-apply.
- **`TestImport_BundledSkillFromClaude`** (`internal/cli/import_test.go`) — fixture: `SKILL.md` + `scripts/deploy.sh` (0o755) + `references/notes.md` + `assets/logo.png` + `references/sub/extra.md`. Asserts every relative path + bytes round-trips to `.agentsync/skills/deploy/…` and `scripts/deploy.sh` retains mode `0o755`.
- **`TestRoundTrip_Command_ByteStable`** (`internal/adapter/roo/ingest_test.go`) — render command to disk, read bytes, ingest, re-render, assert identical bytes.
- **`TestMCPAdd_ReadsBackEnvAndAgents`** (`internal/cli` mcp add tests) — `mcp add` with `--env K=v --agents claude,codex`, then load canonical and assert `Env["K"]=="v"` and the agents list round-trips.
- **`TestRestore_PreservesUntrackedFile`** (`internal/git/git_test.go`) — commit fixtures, drop an untracked file, `Restore` to a prior checkpoint, assert the untracked file still exists with identical bytes.
- **`TestRestore_PreservesModeBit`** (`internal/git/git_test.go`) — commit a file, re-commit it with `+x`, `Restore` across the change, assert `Perm()` matches the target checkpoint.
- **`TestStage_WriteDriven`** (`internal/git` or `internal/cli`) — assert every `written` member appears in the checkpoint tree, including an identical-content rewrite and a mode-only change.
- **`TestApply_InitOverNonEmptyDir`** (`internal/cli`) — apply into a dir with pre-existing unmanaged files; assert repo init + first checkpoint + pre-existing files untouched.
- **`TestRevert_PreservesUntouchedTrackedHandEdit`** (`internal/cli/revert_test.go`) — apply, hand-edit an untouched tracked file, revert, assert the edit survives as a snapshot commit reachable in history.
- **`TestRevert_SharedDirRollsBackAllOwners`** (`internal/cli/revert_test.go`) — apply two owners into a shared root, revert, assert both owners' files roll back and the cross-agent warning prints.
- **`TestRevert_FoldedChildViaParent`** (`internal/cli/revert_test.go`) — folded child root prints the parent-revert notice and is rolled back via the parent.
- **`TestConcurrentRevert_LockSerializes`** and **`TestConcurrentRevert_DryRunDoesNotContend`** (`internal/cli/revert_test.go`) — mirror the apply concurrency tests.
- **`TestRevert_Errors/all_plus_to`** (extend `TestRevert_Errors`) — assert `--all`+`--to` is rejected.
- **`TestRevert_ToNonAncestorRejected`** (`internal/cli/revert_test.go`) — a resolvable-but-non-ancestor `--to` ref is rejected (or `t.Skip` referencing the code-fix issue if that lands separately).
- **`TestApply_ResolvedSecretInHistoryNotInSource`** (`internal/cli`) — `${secret:K}` resolves into the destination checkpoint (cleartext present in git history) while the canonical source retains `${secret:K}`; no dest→source write bypasses `capture.Capture`.
- **Carried-over adapter shards** — `TestRoundTrip_<X>_OnDisk` per package (OpenCode mode/frontmatter; Codex skill-bundled-files/body/MCP+hooks/subagent-name; Gemini subagent/hook-group/namespaced-commands; Cursor Extra; generic sse→http; Windsurf workflow + `trigger: glob` memory; Cline workflow), each anchored to a spec-complete on-disk fixture with the unmodeled key / bundled file / non-default trigger, asserting on-disk bytes survive.

---

## 8. Docs to update

- **`docs/capability-matrix.md`** (mirrored to the website at build time) — for **every** cell whose support claim a conversion now actually backs or corrects (e.g. a `◐` that should be a `✓`/`✗`, or vice-versa, once the artifact-anchored test proves what truly round-trips). Per CLAUDE.md, a capability claim must point to the test that backs it — these conversions are those tests.
- **`README.md`** + **`docs/user-guide.md`** matrices — mirror any capability-matrix cell changes in the same commit.
- **`docs/architecture.md`** / **`docs/components.md`** — only if a conversion reveals that a documented render/capture contract sentence is now false; fix it in the same commit (treat a stale doc as a bug).
- If the `revert --dry-run` lock-skip is documented as a known limitation rather than fixed, note it in the relevant CLI reference (`docs/user-guide.md` command reference / `website/src/content/docs/reference/cli.mdx`).
- Otherwise **no doc change is required for the pure test scaffolding** that does not move a capability claim — but the matrix/README/user-guide updates above are non-negotiable for any cell whose truth value a conversion establishes.

---

## 9. Out of scope / do not touch

- **The underlying product code fixes** — #124/02/08/09/10/11/13/14/16 and the `revert --to` ancestor-validation / dry-run-lock code changes are tracked in their own issues. This issue lands the **tests**; depend on those fixes, do not re-implement them here. If a conversion surfaces a *new* product bug, file it separately and `t.Skip` with a reference rather than asserting buggy behavior.
- **The QA / manual passes** #164 / #167 / #173.
- Do not weaken the secret-handling invariants: no new dest→source write outside `capture.Capture`; no new secret-bearing field outside `walkSecretFields`; do not unwrap `secrets.Resolved` outside an adapter `Render`.
- Do not change the canonical schema or the `Adapter`/`DestWriter` contracts.
- Do not introduce testify/gomega or `os.UserHomeDir()` in tests.

---

## 10. References

- Review report sections: **C1**, **E#18** (`E18-C1-systematic-conversion`), **§F bullet 4** (`F4-bidirectional-restore-unverified`), and the per-area items `D3-claude-lsp-hook-model-anchored-tests`, `D2-import-bundled-skill-untested`, `D2-mcp-add-tests-no-readback`, `D3-roo-command-model-seeded`, `D2-apply-no-shared-dir-revert-e2e`, `D2-apply-untouched-tracked-edits-uncommitted`, `D2-apply-stage-skips-unmodified`, `D2-restore-mode-bit-fidelity-untested`, `D2-revert-lock-serialization-untested`, `D2-revert-dryrun-unsynchronized-index-read`, `D2-revert-all-plus-to-untested`, `D2-revert-to-ref-no-ancestor-check`.
- Key files:
  - `internal/adapter/claude/ingest_test.go:229` (`TestIngest_RoundTripsHooksButIgnoresSettingsLSP`)
  - `internal/cli/import_test.go:670` (`TestImport_SkillFromClaude`)
  - `internal/adapter/roo/ingest_test.go:20` (`TestRoundTrip_ProjectScope`)
  - `internal/cli/mcp.go` (mcp add tests)
  - `internal/git/git_test.go:517` (`TestRestoreAppendOnly`)
  - `internal/git/restore.go:67-102` (`Restore`, the two-reset path)
  - `internal/git/commit.go:18-80` (`Stage`, `StageTrackedDeletions`, `CommitStaged`, `SnapshotDirtyTracked`)
  - `internal/git/log.go:56-74` (`Resolve`, `HasMultipleCheckpoints`)
  - `internal/cli/gitbackup.go:80-106` (checkpoint staging), `:141-183` (shared-root owners)
  - `internal/cli/revert.go:45-74` (arg guards + lock), `:172-253` (`revertRoot`), `:256-275` (`previewRevert`)
  - `internal/cli/revert_test.go` (existing revert tests to mirror/extend)
  - `internal/adapter/generic/homedir_test.go:60-89` (`TestVersionRoots_AllSpecs`)
- Governing invariant: **CLAUDE.md → "Models must stay faithful to their on-disk artifacts — non-negotiable"** (the "Fidelity tests anchor to the artifact, not the model" rule) and the **docs-sync** + **test-conventions** sections.

---
📋 _Part of the v1.1 remediation epic #178 — see it for the full wave plan and dependency graph._

spxrogers:
## Summary

Wave 4 of the agentsync v1.0 remediation epic (#178) — the cross-cutting hardening
layer: shared Render-time and ingest-read-error guards, the per-adapter low/nit
backlog across all nine deep adapters + the generic tier, the core/git nit batch,
the CLI-UX polish batch, the artifact-anchored test-fidelity sweep, and the QA/fuzz
verification passes. All work is on branch `claude/agentsync-wave-4-0nq1ay`.

Closes: #153, #156, #159, #164, #167, #173, #170, #176, #151, #155, #158, #162,
#166, #169, #172, #175 (all 16 Wave-4 sub-issues).

Local gates: `AGENTSYNC_TEST_IN_CONTAINER=1 go test -count=1 ./...` green (29 pkgs ok),
and `golangci-lint@v2.12.2` reports 0 issues (run with `GOTOOLCHAIN=go1.26.2`); no
gofmt/gofumpt/`go mod tidy` drift.

### How each sub-issue's acceptance criteria were met

**#176 fix(core): carried-over core/git nits** — `Hook.Event` is now
`untrusted.Text` (raw via `.Unverified()` at machine/FS sites, sanitizing
`.String()` at display sites); one shared `adapter.DispatchOps` replaces the ten
duplicated `Apply` dispatches (continuedev's direct `w.Write` and every error-wrap
string preserved; `noop` untouched); `Registry.Names()` uses `sort.Strings`;
`FileOp.Action` doc enumerates `""` as write; the `AtomicWrite` fsync comment now
matches `state.Save`; `git.Detect` fails safe to `StateForeign` on a non-`ErrRepositoryNotExists`
error with a "meaningful only when err==nil" note; `Plan`'s switch handles
`merkletrie.Modify` explicitly; the per-apply git-status cost is short-circuited for
an owned root with no writes/deletions; `shortStr/shortHash/shortRef` collapse to one
`git.Short`; the skipped-init sticky-`mode=on` note prints. Tests: `TestLoadHooks_EventIsUntrusted`,
`TestDispatchOps`, `TestNames_Sorted`, `TestDetect_ErrorReturnsForeign`,
`TestPlan_ModifyAndUnknown`, `TestShort`.

**#156 fix(render): shared Render-time ValidateComponentID guard** — `render.Plan`
validates every component id (`Subagent`/`Command`/`Skill` Name) via
`source.ValidateComponentID` and enforces a `filepath.Clean` containment backstop
over every emitted `FileOp.Path`, at the single dispatch waist (no per-adapter
edits). Names reach `Plan` via a new string-only `secrets.Resolved.ComponentIDs()`
accessor, so the `Resolved.Canonical` forbidigo fence stays intact (no `//nolint`).
Tests: `TestEveryAdapterRejectsTraversalComponentName` (all 31 adapters × 8 malicious
names × both scopes, 559 subtests), `TestPlan_RejectsTraversalComponentName`,
`TestPlan_ContainmentBackstopRejectsEscapingFileOp`. Docs: `architecture.md` §7,
`components.md`, `CHANGELOG`.

**#159 fix(adapters): harmonize ingest read-error handling** — shared
`adapter.ReadFileOptional`/`ReadDirOptional` (absent → skip; any other read/parse
error → surfaced) applied across all ten deep adapters + the generic tier; Claude's
MCP-loud / hooks-silent asymmetry removed. Per-adapter `TestIngest_Unreadable*` tests
(directory-at-path / file-at-dir for root-proof unreadable fixtures). Read-side only.
Docs: `architecture.md`, `CHANGELOG`.

**#155 fix(cli): status/diff/reconcile/apply/mcp/doctor UX + stream separation** —
all 15 items: reconcile prompt/`[d]iff` show masked values (not SHA prefixes); apply
reports deletions in the headline; `status --exit-code`/`diff --exit-code` (code 2);
doctor resolves `${secret}`/`${env}` refs; interactive prompts to stderr; `diff`
distinguishes unmanaged paths + gains `--agents`; reconcile persists dest-removed MCP
deletions through the approved funnel; `mcp add --header` for http/sse; import lists
bundled skill files; split-stream test helper; real lock-contention test. No
`walkSecretFields` change, no unguarded dest→source write. Docs: `user-guide.md`,
`README.md`, `cli.mdx`, `CHANGELOG`.

**#151 fix(codex): dead OwnedKeys + MCP/hook Skips** — removed the dead hand-set
`renderHooks` `OwnedKeys` (render.Plan populates it from state); `renderMCP` now
returns `[]adapter.Skip` and emits a `SkipDropped` per disabled/non-targeted server;
a non-command hook handler emits a `SkipReduced` yet still renders (Codex tolerates
any type). `TestMergeTOML_RemovesOrphanedHookKey`, `TestRenderMCP_SkipsDisabledAndNonTargeted`.
Skip guards stay green. Capability matrix unchanged (reporting improvement, not
coverage). `CHANGELOG`.

**#166 fix(gemini): hook group shape, namespaced commands, $-escape** — renderHooks
coalesces consecutive `(event,matcher)` Hooks into one native multi-handler group
(idempotent); empty `type` omitted; always-fire-event matcher dropped with a
`SkipReduced`; commands round-trip through subdir namespacing (`git/commit.toml` ↔
`/git:commit`, scoped to the adapter native round-trip since the flat canonical
source can't carry `/`). Verified upstream that Gemini's `envVarResolver` has **no**
`$`-escape, so a resolved value containing `$` is written verbatim and flagged with a
`SkipReduced` rather than corrupted (destination corruption, not a cleartext leak).
5 artifact-anchored tests. Docs: `capability-matrix.md` (Gemini), `CHANGELOG`.

**#169 fix(windsurf): global_rules guard + doc notes** — the two memory reads route
through `readMemoryBody` with an else-if guard so a populated project rule can't be
clobbered by a stray global read (reconciled with #159 so it still surfaces a real
read error loudly); documented the `url→serverUrl` canonicalization, the ~12,000-char
workspace-rule/workflow limit, and the `.devin/rules` rebrand in code + the matrix.
Artifact-anchored whole-file-overwrite/backup and frontmatter-carrying round-trip
tests. `CHANGELOG`.

**#172 fix(continue+cursor): MCP round-trip + off-spec key** — Continue round-trips
the block `name/version/schema` header (reserved `Extra` keys) instead of regenerating
`0.0.1`/`v1`, and reports a Skip for a `command`+`url` ambiguity instead of silently
dropping the URL; Cursor stops emitting the off-spec `type` key on remote (`url`-bearing)
servers (stdio keeps it; ingest stays tolerant; `TODO(#164)` for the reject-vs-ignore
verdict). Continue/cursor full-cycle Extra, orphan-cleanup, precise drop-detection,
`is_background` tests. Docs: `capability-matrix.md`, `CHANGELOG`.

**#175 fix(adapters): roo/cline/generic transport + Detect, noop, claude int precision** —
Roo untyped-url MCP normalizes to `http` (stable, documented) + Detect drops the
phantom `roo` binary probe + capability/skip correspondence test; Cline workflow
ingest is ownership-marker scoped (no foreign capture); generic keyless `sse→http`
documented + `copilot`/`jetbrains` flagged undetectable (`trae` has a real `.trae`
home marker, so it is NOT flagged — correcting the brief per the upstream-truth rule);
Claude MCP+Hooks `Extra` decode uses `UseNumber` so integers > 2^53 survive; noop
package doc/name-guard + 6 method tests incl. "Apply ignores ops"; VersionedDirs /
HasNestedRepoBelow cost/casing doc comments; matrix per-row transport notes (Roo/Cline/
generic only). Reconciled with #159 and #176.

**#162 fix(core+opencode): skill-file prune, strict TOML, mode-drift, tamper guard** —
all 12 items (A–L): `WriteSkill` prunes orphaned bundled files (idempotent re-import);
component TOML loaders strict-decode (a typo'd key errors; `Extra` map keys still
tolerated; `loadMarketplaces` tolerates the documented `head_sha`/`name` cache keys —
empirically verified go-toml/v2 `DisallowUnknownFields` coexists with a named `Extra`
map); convergence compares `op.Mode` so a content-identical chmod re-converges and
shows as drift; `LoadProjected` reads tree-hash and bodies through the same fs;
`resolveInstalledEntry` memoizes the marketplace scan; `state.Save` self-heal
documented; OpenCode memory fragments round-trip by design (reversible markers +
`CollapseMemoryMarkers`, documented), and the `.opencode/skills → shared ~/.claude/skills`
choice documented (OpenCode scans it — verified upstream); artifact-anchored opencode
ingest tests; MergeJSONC test strengthened. Docs: `architecture.md`, `capability-matrix.md`,
`CHANGELOG`.

**#158 test(release): ldflags-aware version-injection e2e** — `TestRoot_VersionInjection`
sets distinctive `cli.Version/Commit/Date` sentinels and asserts all three surface on
both `--version` and the `version` subcommand (a real oracle: verified red when a field
is dropped from the template). Test-only.

**#170 chore(docs): build-time link/anchor checker + rollback guide** —
`website/scripts/check-links.mjs` validates in-site routes + heading anchors at build
time (wired into `predev`/`prebuild` after `sync:docs`; `github-slugger`), with
`node --test` coverage (slugger fixture + checker-behavior temp-tree test); a new
`guides/rollback.mdx` wired into the sidebar and cross-linked from `cli.mdx`. One
pre-existing #145-owned anchor is allowlisted with a `TODO(#145)`. Website-only.

**#173 qa(release): reproducibility diff** — verified via two goreleaser snapshots and
a `dist/` diff (see the QA note below); the `.goreleaser.yaml` reproducibility claim
reconciled with observed reality.

**#153 test(C1): artifact-anchored fidelity sweep** — the bulk of the sweep was already
satisfied by waves 1–3 and the Phase-3 work (per-adapter `*_SpecComplete`/`ArtifactAnchored`
round-trips, `TestRestore_PreservesUntrackedFiles`/`PreservesFileMode`,
`TestResolvedSecretLandsInLocalHistory`, shared-dir/folded-child revert, etc.). This wave
closes the remaining named gaps: `TestStage_WriteDriven` (checkpoint captures the written
set incl. identical-content rewrite + mode-only chmod), `TestMCPAdd_ReadsBackEnvAndAgents`,
`TestImport_BundledSkillFromClaude` (bundled `scripts/`/`references/`/`assets/` + 0o755 bit),
`TestConcurrentRevert_LockSerializes`/`DryRunDoesNotContend`, `TestRevert_Errors/all_plus_to`
(the non-ancestor `--to` subcase is a `t.Skip` — ancestor-validation is not implemented in
`log.go`/`revert.go`, so asserting the current accept-any-resolvable-commit behavior would be
asserting the unsafe path), roo `TestRoundTrip_Command_ByteStable`, and two `// MODEL-ANCHORED:`
annotations. Test-only, all green.

**#164 qa: git-backup/revert + untested adapters** — verification pass (see the QA note
below). The data-loss findings are reproduced-and-fixed by existing regression tests; the
edge cases it flagged are already handled; real-harness live-testing is not possible in the
hermetic container, so adapters are covered by artifact-anchored round-trips instead.

**#167 qa(secrets): backstop + .git-perm fuzz** — `TestResidualSecretCleartext_Adversarial`
(both prongs, length boundary, Extra passthrough incl. project scope, nil guards),
`FuzzResidualSecretCleartext` (value-leak safety property, 30s / 330k execs, 0
counterexamples), locked-backend + EnvBackend edge tests, `CheckIdentityPermissions`
coverage, `.git` perm lifecycle (`Init` 0o700 → loosen → `EnsureDirSecure` re-tighten/warn),
`SnapshotDirtyTracked` cleartext-capture. Test-only. Routed findings below.

---

### #164 / #167 QA verification notes

**#164 — git-backup/revert & adapters.** In this hermetic, offline container the seven
"untested-in-the-wild" harnesses (Cursor, Continue, Windsurf, Roo, Cline, Gemini, generic)
cannot be launched, so the F2 "live real-harness load" step is **logged as not coverable
here**; those adapters are instead exercised by artifact-anchored render→ingest→re-render
round-trips. The F1 git data-loss findings are reproduced-and-fixed by existing tests:
untracked survival (`TestRestore_PreservesUntrackedFiles`, `TestRevert_PreservesUntrackedUserFiles`),
first-apply-revertible (`TestApply_GitBackupBaselineRevertsFirstApply`), pre-existing
content (`TestApply_GitBackupBaseline{RestoresPreexistingManagedFile,PreexistingContentOutOfContract}`),
symlinked-foreign-repo detection (`TestHasNestedRepoBelowFollowsSymlink`), and
nested-foreign-repo-under-owned-root revert refusal (`TestRevertRootSkipsForeignNestedRepo`).
The Windows `.git` chmod no-op is already documented in README Known-limits.

**F5 partial-downgrade re-read verdicts:** (1) OpenCode subagent/command round-trips **are
now artifact-anchored** (`TestRender_*_PassthroughOrSkip_OnDiskFixture`, `TestRoundTrip_OpenCode_SpecComplete`,
#162 ingest artifact tests) — adjust. (2) Roo command byte-stability — **addressed** by #153's
`TestRoundTrip_Command_ByteStable`. (3) reconcile `[d]iff` SHA-prefix-only — **fixed** by #155
(masked values). (4) locked-backend capture refusal — **confirmed** warning-only degradation of the
VALUE prong (SLOT prong still fires); routed to #135/#163. (5) Windsurf round-trips — **now
artifact-anchored** (#169). (6) `.git` chmod-failure framing — **confirmed conditional** (POSIX-only,
Windows no-op). (7) `--to` ancestor-validation severity — **confirmed low** (bounded/append-only-recoverable).
(8) revert no-confirmation — **confirmed low** (append-only recoverability). The Cursor remote-`type`
verdict feeds #172 (drop remote `type`; `TODO(#164)` marker for the reject-vs-ignore live check).

**#167 routed findings (not fixed here — test-only):** (a) under a **locked** secrets backend the
capture backstop's VALUE prong goes silent (no live values resolve), so a genuine value-leak is not
refused by that prong — the backend-independent SLOT prong still fires; a hardening (source the detection
set independent of the live backend) is a candidate for #135/#163. (b) `runtimeIsWindows`/`os.Getenv`
are read directly (no injectable seam), so the Windows no-op branches of `CheckIdentityPermissions` and
`tightenGitDir` cannot be unit-forced on Linux — a `var runtimeIsWindows` seam would close that coverage
gap.

**#173 reproducibility:** two snapshot releases built from the same commit with the
release-pinned goreleaser 2.16.0 were diffed. Verdict: the binaries and all six tar.gz/zip
**archives are byte-identical** run-to-run (confirming #138's restored mtime-pin), but
`checksums.txt` is **NOT** — the four nfpms deb/rpm lines differ every run because the packages
embed the wall-clock **build** time (root-caused to the deb `data.tar` `./usr`+`./usr/bin`
directory-entry octal mtimes and the rpm `RPMTAG_BUILDTIME` + its signature digests; the packaged
binary keeps the pinned commit mtime). The `.goreleaser.yaml` "(reproducible, identical)"
overclaim was corrected (comment-only — making the packages reproducible needs an `nfpms.mtime`
pin, a behavioral change out of #173's QA scope, noted for #138), and a guard test
(`internal/release/reproducibility_test.go`) pins the archive mtime-pin against the PR #115
regression class.

---

### Note on commit verification

Commits carry the correct `noreply@anthropic.com` identity but show GitHub's "Unverified"
badge: the container has no commit-signing key provisioned (`/home/claude/.ssh/commit_signing_key.pub`
is a 0-byte file, no private key), so `git` cannot produce a signature. This is an environment
provisioning gap, not a code issue.

## Type of change

- [x] Bug fix
- [x] Refactor (no behavior change)
- [x] Docs
- [x] Tests / CI / tooling

## Test plan

- [x] `AGENTSYNC_TEST_IN_CONTAINER=1 go test -count=1 ./...` green (equivalent to the
      container `just test` — podman isn't available in this environment, so the suite
      runs directly with the container env gate the script sets).
- [x] `GOTOOLCHAIN=go1.26.2 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run ./...`
      is clean (0 issues); no gofmt/gofumpt/`go mod tidy` drift.
- [ ] `just test-release` / `just lint` — not runnable here (no podman/`just` binary); the
      equivalent underlying commands above are green.

## Checklist

- [x] Conventional commit messages with a scope.
- [x] Tests added/updated for the behavior changed (artifact-anchored where the issue required it).
- [x] Touches `internal/secrets`/`internal/capture`/`source.Write*`: re-read the secret-handling
      invariants; no `walkSecretFields` change, no dest→source write outside `capture.Capture`,
      forbidigo fence + fail-closed backstop intact.
- [x] Docs updated (same commit) where behavior, CLI surface, or capability coverage changed.

https://claude.ai/code/session_01Qz79FfYbSCJWuyHvsyb485

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qz79FfYbSCJWuyHvsyb485)_


spxrogers:
## Post-review hardening — four-lens review loop (converged)

After opening this PR I ran a four-lens adversarial review loop (correctness /
adversarial / api-design / test-rigor, in parallel each round) over the Wave-4
diff. It converged at **Round 3 with four-lens unanimity** (`CLEAN — ship it`).
Three follow-up commits are on the branch:

**Round 1 — `f12e144`** (2 BLOCKER-grade, 1 ISSUE, NITs — each break-verified):
- 🔴 **Security:** the Render-time path-traversal guard validated only
  subagent/command/skill names, **not MCP/LSP server ids** — yet continuedev joins
  an MCP id into `filepath.Join(MCPDir, id+".yaml")` and a marketplace-projected id
  is an untrusted raw manifest key. A plugin's `mcpServers` key like `../../../etc/x`
  was a write-anywhere primitive. `ComponentIDs()` now yields MCP/LSP ids and `Plan`
  validates them (once, model-wide).
- 🔴 **Correctness:** the merge-toml-keys state recorder had a guard
  (`assertHashStableTOMLValue`) that **failed apply on any numeric owned value**
  (e.g. a Codex MCP `timeout`). Removed — an in-range int hashes identically as the
  JSON source and TOML dest, so there was no real drift to guard.
- 🟠 **Fidelity:** `MergeExtra` leaked `__`-prefixed agentsync-internal metadata
  (continuedev's block headers) cross-adapter into other agents' native configs.
  Fixed with a reserved-prefix skip.
- NIT: hoisted id-validation out of the per-agent loop; agent-agnostic error.

**Round 2 — `863a34d`** (no BLOCKER/ISSUE; hardening + doc/test rigor):
- Made the `__` reserved-`Extra` namespace **symmetric** — `ExtraNativeKeys`
  (capture) now refuses to ingest `__` keys, matching `MergeExtra` (render), so a
  stray native `__` key can't be captured-then-silently-dropped.
- Broadened the stale `ComponentID` struct doc; tightened the numeric
  state-recording test to assert the exact pointer + a real no-false-drift hash
  equality; added direct `ComponentIDs()` coverage across all five kinds + the
  project overlay. Documented the `__` namespace in architecture §8 + components.md.

**Round 3 — `4a7dc18`** (converged; one framing-honesty NIT):
- Reworded the project-overlay id-recursion comment + two test docstrings: since
  `project.Merge` (`overlayByKey`) already mirrors every project id into the
  top-level model, the recursion is forward-looking defense-in-depth, not the sole
  production guard — now stated faithfully (per the repo's "doc claims are not
  self-certifying" rule).

**Deferred:** scoping id-validation to only servers that actually render (a disabled
or non-targeted server with a malformed id currently fails the whole plan) is a
behavior change, tracked separately in **#197** rather than expanding this PR.

Docs updated in the same commits (architecture §7–8, components.md, CHANGELOG #156).
Full suite green + `golangci-lint` 0 issues after each round.

---

Issue #248, opened 2026-09-04
render.Writer.Write chmods through a symlinked destination before the symlink policy is consulted


spxrogers:
With `AGENTSYNC_ALLOW_SYMLINK_DEST` unset, `apply` is documented as refusing to write through a symlinked destination. The refusal lives in `iox.AtomicWrite` (`resolveSymlinkDest`), which `render.Writer.Write` reaches only when the content differs. Before that, `Write`'s convergence arm handles the content-identical case itself: it reads through the link, and when `op.Mode != 0` and the permission bits differ it calls `os.Chmod(op.Path, want)`. `os.Chmod` follows symlinks, so the TARGET's mode is rewritten through a link the user never opted into.

**Measured on `7f2d4ad`** (probe in a throwaway clone): link → regular file with identical content, target 0600, `op.Mode` 0644, env unset → `Write` returns nil, the link is preserved, the target is now 0644. No `ErrSymlinkDest`.

**Where.** `internal/render/writer.go`, the convergence branch (`os.ReadFile(op.Path)` then the `op.Mode != 0` chmod) runs before `AtomicWrite`; `resolveSymlinkDest` is never consulted on that path.

**Why it matters now.** #247 makes every read-side surface obey the switch, and `reconcile`'s `[o]verride` re-applies through `Writer.Write`. `[o]` is therefore withheld on the symlink refusal arm in #247, and the SECURITY/architecture prose there names this issue as the one write path that still follows a link with the switch unset. The same convergence read is also the unguarded read behind #241 (a FIFO at a rendered destination hangs `apply`), so a fix that routes the convergence arm through the same shape and symlink gates would close both.

**Expected.** With the switch unset, `Write` refuses a symlinked destination on the mode arm exactly as `AtomicWrite` refuses it on the content arm (or skips the chmod and reports it), so "refuses to write through a link" is true of every write.

Issue #242, opened 2026-09-01
import &lt;agent&gt; hangs forever on a non-regular destination (all adapters)


spxrogers:
Found during review of #240, which fixes the same defect class for `status` / `diff` / `reconcile`'s drift walk. This one is in the adapter `Ingest` paths and is **not** fixed by that PR.

`os.ReadFile` on a FIFO does not fail — it blocks in the open — so the read's own error path never runs and the command never returns.

## Measured

Built at `d4fb233` (i.e. **with** #240's gate in place), clean apply, destination replaced by a `0600` FIFO, neutral cwd:

```
import claude              rc=124  *** HUNG ***
import claude --dry-run    rc=124  *** HUNG ***
```

Reproduced with a FIFO at a whole-file destination (`~/.claude/skills/demo/SKILL.md`) and at a key-merge one (`~/.claude.json`).

Note `agentsync import` with **no** agent is `cobra.ExactArgs(1)` and fails instantly with `accepts 1 arg(s), received 0` — a selector is required to reach the bug, and forgetting one is how an earlier test of this silently measured nothing.

## `doctor` is exposed through the same path

`doctor` performs no destination read of its own, but its plugin check reaches one: `checkPlugins` → `undeclaredNativePlugins` (`internal/cli/plugin_drift.go`) → `claude.IngestPlugins` → a bare `os.ReadFile(p.Settings)` at `internal/adapter/claude/ingest_plugins.go:35`. `settings.json` is a managed destination (claude renders `/hooks` and `/lspServers` into it).

A reviewer measured `doctor` hanging at rc=124 with a FIFO there, wedged mid-"Plugins". A second fixture exited rc=1 instead, because an unrelated issue was reported before the plugin check ran — so reachability depends on what else `doctor` finds first, but **the unguarded read is there by inspection** and `doctor` should be treated as in scope for this issue rather than immune.

## Where

Two distinct read paths, both unguarded:

1. **`internal/adapter/readoptional.go:25`** — `ReadFileOptional`'s `os.ReadFile`, reached from `claude.(*Adapter).Ingest` ← `cli.importRun`. 27 call sites across the adapters.
2. **31 direct `os.ReadFile` calls** in the adapter packages that do not go through `ReadFileOptional` — e.g. `internal/adapter/claude/ingest.go:82`, `ingest_plugins.go:35`, and the equivalents in `codex`, `cursor`, `gemini`, `generic`, `windsurf`, `roo`, `cline`, `continuedev`.

[...]

Issue #241, opened 2026-09-01
apply and apply --dry-run hang forever on a non-regular destination


spxrogers:
Found during review of #240, which fixes the same defect class for `status` / `diff` / `reconcile`'s drift walk. This one is in `internal/render` and is **not** fixed by that PR.

`os.ReadFile` on a FIFO does not fail — it blocks in the open waiting for a writer that never comes — so the read's own error path never runs and the command never returns.

## Measured

Built at `d4fb233` (i.e. **with** #240's gate in place), clean apply, then the rendered destination `~/.claude/skills/demo/SKILL.md` replaced by a `0600` FIFO, run from a neutral cwd:

| command | rc |
| --- | --- |
| `status` | 0 |
| `diff` | 0 |
| `reconcile --auto-safe` | 0 |
| **`apply --dry-run`** | **124 — hung** |
| **`apply`** | **124 — hung** |
| **`reconcile --auto-override`** | **124 — hung** |

`apply --dry-run` is advertised as read-only.

`reconcile --auto-override` (and the interactive `[o]` keystroke) belongs here rather than in #240: `[o]` re-applies the item through `render.NewWriter` → `Apply` → the same unguarded read, so it shares `apply`'s bug rather than the drift walk's. #240 therefore withholds `[o]verride` from `writeBackFileItem`'s refusal message — recommending it would route a user out of a clean refusal into this wedge. **When this issue is fixed, that suggestion can be restored** (there is a test asserting the message does not currently offer it, which will need updating in the same commit).

## Where

`internal/render/writer.go:200` — the convergence read, the first statement of `Writer.Write`:

```go
if cur, err := os.ReadFile(op.Path); err == nil && bytes.Equal(cur, finalBytes) {
```

Its own comment notes "The same read drives the dry-run preview: it is what lets `apply --dry-run` label a converged destination 'synced' rather than 'write'." So both the real apply and the preview go through it.

## A false doc claim to fix alongside

`internal/render/writer.go`, the doc comment on `isRegularOrAbsent`, currently says:

> … without this, `apply --dry-run` (advertised as read-only) and the real apply's pre-delete read would both hang forever on a FIFO left at a destination path.

The predicate exists and is correct, but `Writer.Write`'s convergence read does not call it, so the claim about `apply --dry-run` is false. Whatever fixes the hang should make that sentence true rather than merely plausible.

## Suggested shape

[...]

PR #265, opened 2026-09-11
fix(render): keep shared dests; refuse FIFO and symlink dest writes

### LINKED ISSUE #246: apply can delete a shared destination that another enabled agent still renders (order-dependent)
`applyPlan` computes each agent's orphan deletes from that agent's ops alone and never excludes a path another enabled agent still renders. `reconcile` performs exactly that exclusion (the `rendered` set in `collectReconcileItems`); the apply path has no counterpart. What happens to the file then depends on registry order.

**Where.** `internal/render/pipeline.go` calls `orphanDeletes(st, userHome, name, scope, project, res.Ops)` per agent inside `for _, name := range reg.Names()`; `internal/render/state_apply.go`'s `orphanDeletes` builds its `rendered` set from the passed ops only. The run-scoped `deletedOrphans` map dedupes a path across agents but says nothing about whether anyone still renders it. `reg.Names()` is sorted.

**Measured on `7f2d4ad`.** Two agents own `~/.claude/skills/demo/SKILL.md` in state (Claude and OpenCode genuinely share `~/.claude/skills/`; many breadth-tier agents share `.agents/skills`). With the dropping agent passed no ops and the keeper passed the write op, each direction synthesizes one delete for a path the other agent still renders.

- **Dropper before keeper** (e.g. `claude` drops): the delete runs, the keeper rewrites the file. It survives, at the cost of a spurious delete plus rewrite on every apply. `PruneStaleState` then keeps the dropper's state entry because `Lstat` finds the keeper's rewrite, so the dropper also shows a permanent `orphan` row in `status`.
- **Dropper after keeper** (e.g. `opencode` drops): the keeper writes, the dropper deletes, and then `RecordOpsState(keeper)` fails on `os.ReadFile(op.Path)` for the now-missing file. `apply` exits non-zero and state is never saved, so the identical sequence repeats on every run as a hard apply error. The component never lands.

**Reachable configurations.** Plugin targeting (`source.PluginSpec.Agents`) narrows which agents receive a plugin's components; `native_agents` (#223) removes an agent from projection outright. Either makes one agent stop rendering a shared skill a sibling still renders. An adapter that starts emitting an `adapter.Skip` for a component it used to render produces the same shape.

**Expected.** `apply` excludes from an agent's orphan deletes any path another enabled agent still renders in the same run, the rule `reconcile` already applies, or defers deletes until after every agent's writes.

Noticed while verifying #229 axis 15. Out of scope there: it is on the apply/write path, not the shared drift walk.

harryy2510:
Fixes #246, #248, and #242 (and the apply-side FIFO hang in #241, same Writer.Write path as #248).

### #246 — shared dest orphan delete
`apply` computed orphan deletes per agent and never excluded a path another enabled agent still rendered. Codex/Factory/agy sharing `.agents/skills` could delete-then-rewrite or delete-then-fail depending on registry order.

Apply now collects whole-file write paths from the full plan (`StillRenderedWholeFiles`) and skips those deletes. `removalCounts` / `baselinePaths` use the same filter.

### #248 — chmod through a symlink
`Writer.Write`'s content-identical arm called `os.Chmod` on the dest path, which follows symlinks even when `AGENTSYNC_ALLOW_SYMLINK_DEST` is unset. Chmod now goes through `iox.ResolveSymlinkDest`.

### #242 / #241 — FIFO dest hangs
`os.ReadFile` on a FIFO blocks in `open(2)`. Shape-check before open:

- `adapter.ReadFileOptional` / `ReadExisting` (ingest + merge reads)
- `Writer.Write` via `isRegularOrAbsent`

Unskipped the ready-made `import claude` / `apply` rows in `dest_fifo_e2e_unix_test.go`. Remaining `reconcile --auto-override` skip is still #241's override path if it does not go through `Write` the same way; apply itself is covered.

### Tests
Container: `TestFilterOrphanDeletes_KeepsSharedPath`, `TestWrite_ChmodRefusesSymlinkDest`, `TestReadFileOptional_FIFODoesNotBlock`, `TestCommandsDoNotHangOnNonRegularDestination`.

### DIFF
--- a/internal/adapter/claude/apply.go
+++ b/internal/adapter/claude/apply.go
@@ -3,7 +3,6 @@ package claude
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 
 	"github.com/spxrogers/agentsync/internal/adapter"
 	"github.com/spxrogers/agentsync/internal/jsonkeys"
@@ -25,7 +24,10 @@ func (a *Adapter) Apply(ops []adapter.FileOp, w adapter.DestWriter) error {
 // so we still pass the raw FileOp (carrying op.Content) along.
 func (a *Adapter) applyWrite(op adapter.FileOp, w adapter.DestWriter) error {
 	if op.MergeStrategy == "merge-json-keys" {
-		existing := readJSONFile(op.Path)
+		existing, err := readJSONFile(op.Path)
+		if err != nil {
+			return fmt.Errorf("read %s: %w", op.Path, err)
+		}
 		ours, err := jsonkeys.DecodeObject(op.Content)
 		if err != nil {
 			return fmt.Errorf("parse our payload for %s: %w", op.Path, err)
@@ -40,16 +42,19 @@ func (a *Adapter) applyWrite(op adapter.FileOp, w adapter.DestWriter) error {
 	return w.Write(op, op.Content)
 }
 
-func readJSONFile(path string) map[string]any {
-	data, err := os.ReadFile(path)
+func readJSONFile(path string) (map[string]any, error) {
+	data, err := adapter.ReadExisting(path)
 	if err != nil {
-		return map[string]any{}
+		return nil, err
+	}
+	if len(data) == 0 {
+		return map[string]any{}, nil
 	}
 	// Decode preserving json.Number so a foreign integer > 2^53 isn't rounded
 	// when the merged file is re-marshalled.
 	m, err := jsonkeys.DecodeObject(data)
 	if err != nil {
-		return map[string]any{}
+		return map[string]any{}, nil
 	}
-	return m
+	return m, nil
 }
--- a/internal/adapter/claude/ingest.go
+++ b/internal/adapter/claude/ingest.go
@@ -79,14 +79,12 @@ func (a *Adapter) Ingest(scope adapter.Scope, project string) (source.Canonical,
 				continue
 			}
 			skillDir := filepath.Join(p.SkillsDir, e.Name())
-			data, err := os.ReadFile(filepath.Join(skillDir, "SKILL.md"))
+			data, present, err := adapter.ReadFileOptional(filepath.Join(skillDir, "SKILL.md"))
 			if err != nil {
-				// A directory with no SKILL.md is not a skill — skip silently.
-				// A present-but-unreadable SKILL.md is surfaced (never a silent
-				// drop) so the user can fix it, matching the parse-error warning.
-				if !os.IsNotExist(err) {
-					fmt.Fprintf(warn, "warning: skipping skill %q: read SKILL.md: %v\n", e.Name(), err)
-				}
+				fmt.Fprintf(warn, "warning: skipping skill %q: read SKILL.md: %v\n", e.Name(), err)
+				continue
+			}
+			if !present {
 				continue
 			}
 			fm, body, lenient, err := ParseFrontmatterWithReport(data)
--- a/internal/adapter/claude/ingest_plugins.go
+++ b/internal/adapter/claude/ingest_plugins.go
@@ -3,7 +3,6 @@ package claude
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 	"sort"
 	"strings"
 
@@ -32,12 +31,12 @@ func (a *Adapter) IngestPlugins(scope adapter.Scope, project string) ([]adapter.
 		return nil, nil, err
 	}
 	p := ResolvePaths(a.opts.TargetRoot, project, scope == adapter.ScopeProject)
-	data, err := os.ReadFile(p.Settings)
+	data, present, err := adapter.ReadFileOptional(p.Settings)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, nil, nil
-		}
 		return nil, nil, fmt.Errorf("read %s: %w", p.Settings, err)
+	}
+	if !present {
+		return nil, nil, nil
 	}
 	var top map[string]any
 	if json.Unmarshal(data, &top) != nil {
--- a/internal/adapter/cline/apply.go
+++ b/internal/adapter/cline/apply.go
@@ -3,7 +3,6 @@ package cline
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 
 	"github.com/spxrogers/agentsync/internal/adapter"
 	"github.com/spxrogers/agentsync/internal/jsonkeys"
@@ -24,7 +23,10 @@ func (a *Adapter) applyWrite(op adapter.FileOp, w adapter.DestWriter) error {
 	if op.MergeStrategy != "merge-json-keys" {
 		return w.Write(op, op.Content)
 	}
-	existing := readJSONFile(op.Path)
+	existing, err := readJSONFile(op.Path)
+	if err != nil {
+		return fmt.Errorf("read %s: %w", op.Path, err)
+	}
 	ours, err := jsonkeys.DecodeObject(op.Content)
 	if err != nil {
 		return fmt.Errorf("parse our payload for %s: %w", op.Path, err)
@@ -40,14 +42,17 @@ func (a *Adapter) applyWrite(op adapter.FileOp, w adapter.DestWriter) error {
 // readJSONFile reads and decodes a JSON object file, returning an empty map on
 // any read/parse error. Decode preserves json.Number so a foreign integer > 2^53
 // isn't rounded when the merged file is re-marshalled.
-func readJSONFile(path string) map[string]any {
-	data, err := os.ReadFile(path)
+func readJSONFile(path string) (map[string]any, error) {
+	data, err := adapter.ReadExisting(path)
 	if err != nil {
-		return map[string]any{}
+		return nil, err
+	}
+	if len(data) == 0 {
+		return map[string]any{}, nil
 	}
 	m, err := jsonkeys.DecodeObject(data)
 	if err != nil {
-		return map[string]any{}
+		return map[string]any{}, nil
 	}
-	return m
+	return m, nil
 }
--- a/internal/adapter/codex/apply.go
+++ b/internal/adapter/codex/apply.go
@@ -2,7 +2,6 @@ package codex
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/spxrogers/agentsync/internal/adapter"
 	"github.com/spxrogers/agentsync/internal/jsonkeys"
@@ -32,8 +31,8 @@ func (a *Adapter) applyWrite(op adapter.FileOp, w adapter.DestWriter) error {
 	// coerced to "empty file": MergeTOML would then merge our section into an
 	// empty map and the write would silently drop the user's foreign config.toml
 	// keys. Fail loud instead.
-	existing, err := os.ReadFile(op.Path)
-	if err != nil && !os.IsNotExist(err) {
+	existing, err := adapter.ReadExisting(op.Path)
+	if err != nil {
 		return fmt.Errorf("read %s: %w", op.Path, err)
 	}
 	ours, err := jsonkeys.DecodeObject(op.Content)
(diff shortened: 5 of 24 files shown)


spxrogers:
thanks for the second contribution, @harryy2510 ! 
added this one to the review queue.

spxrogers:
heya @harryy2510 — thanks for this one! I ran it through the repo's standard four-lens [review-loop](https://github.com/spxrogers/agentsync/blob/main/.agentsync/skills/review-loop/SKILL.md) plus my own pass, per the usual 💅 .

short version: **the direction is right and the `Writer.Write` half is solid**, but the #246 fix leaves `status` reporting something apply will never do, #242 is only partly closed, and none of the docs moved with the code. requesting changes, but none of it is a rethink.

### What's strong

- **the FIFO gate is properly pinned.** remove either the `isRegularOrAbsent` check in `Writer.Write` or the `Stat` gate in `ReadFileOptional` and tests go red, including the e2e `apply` / `apply --dry-run` rows you un-skipped. exactly what those skips were waiting for.
- **the #248 chmod fix is correct**, including the dangling-link and opted-in paths. revert it to `os.Chmod(op.Path, …)` and `TestWrite_ChmodRefusesSymlinkDest` catches it via the target-mode assertion 🙏
- **`readJSONFile` surfacing read errors is a real improvement.** before this, a permission-denied `settings.json` was read as "empty" and atomically replaced with just our keys. that's a quiet data-loss bug you closed across five adapters.
- HOME-relative keying in `StillRenderedWholeFiles` matches how state keys are spelled in both user and project scope. checked that one specifically.
- CI is green, merges cleanly with `main`, and the full `./internal/...` suite passes on my end.

### BLOCKER 1 — the agent that stopped rendering the file keeps its state entry forever, so `status` never goes clean

apply now skips the delete (good), but nothing ever retires the dropping agent's state entry:

- `PruneStaleState` (`state_apply.go:~104`) keeps any reclaimable entry whose file is still on disk (the `Lstat` check). the keeper just rewrote that file, so the entry is kept on every run.
- `walkPlanItems` (`planwalk.go:276`) builds its orphan list from an unfiltered `render.OrphanFiles`, so `status` / `diff` list the shared file as `orphan` ("will be deleted") for the dropper. apply never deletes it now.

net effect: `status --exit-code` stays non-zero forever after a clean apply. two reviewers reproduced it in both registry orders. (`reconcile` is fine; `collectReconcileItems` already excludes shared paths.)

the fix is small: have `PruneStaleState` drop an entry whose path is in `StillRenderedWholeFiles`, or apply the same keep-set in `OrphanFiles`. easy to pin, too: extend `TestRenderApply_SharedOrphanKeptWhenSiblingStillWrites` to prune and then assert `OrphanFiles(st, …, "opencode", …)` is empty.

### BLOCKER 2 — docs + CHANGELOG

per CLAUDE.md, docs change in the same commit as the behavior. this PR makes several existing sentences false:

- `CHANGELOG.md:193` and `:243-249` (both under `[Unreleased]`) still say apply, `--dry-run`, `[o]verride` and import "are NOT fixed" and still hang.
- `docs/components.md:522-526`: "not yet true of … `Writer.Write` … still block".
- `docs/architecture.md:1082` and **`SECURITY.md:104-105`** still list #248 as the known gap. the security doc is the one I care most about keeping honest.
- `[Unreleased]` needs Fixed entries for #246, #248 and #241 (and #242 once it's actually covered; see below).
- stale comments: `internal/cli/destread.go:57-63`, `internal/cli/reconcile.go:~1469-1522` (leaving `[o]` out of that advice may still be right, but the reason is now "refuses", not "hangs"), and the `readJSONFile` doc comments in cline / cursor / roo / windsurf, which still say "empty map on any read/parse error". only the parse half is true now.

### ISSUE — #242 is only partly fixed

only claude's `SKILL.md` and settings reads moved to `ReadFileOptional`. about 20 raw `os.ReadFile` calls remain on ingest paths:

- claude `ingest.go:112,142` (subagents, commands).
- the skill / subagent / command loops in codex, cursor, opencode, gemini, generic, windsurf, roo, cline and continuedev.
- `codex/ingest_plugins.go:38`.

`import claude` still hangs with a FIFO at `~/.claude/agents/x.md`, and so does `import codex` with one at `~/.codex/prompts/x.md`. the un-skipped `import claude` e2e row passes only because its fixture puts the FIFO at `SKILL.md`.

either finish the sweep (it's the same mechanical swap you already did for claude) or change the body to "Refs #242" so merging doesn't auto-close it. either works for me.

### ISSUE — the `--auto-override` skip is stale

`[o]verride` goes through `a.Apply` → `readJSONFile` / `Writer.Write`, both guarded now. drop the skip at `dest_fifo_e2e_unix_test.go:118` (and the "second group still hangs" comment above it) and both shapes pass. worth adding a `doctor` row while you're there; it passes too.

### ISSUE — test gaps

a mutation sweep turned up some behavior nothing pins:

| mutation | result |
|---|---|
| any of cline / cursor / roo / windsurf / claude `readJSONFile` swallows the read error again | ❌ unpinned ×5 |
| codex / gemini / generic / opencode `applyWrite` back to `os.ReadFile` | ❌ unpinned ×4 |
| drop `FilterOrphanDeletes` from `removalCounts` (`cli/apply.go:448`) | ❌ unpinned |
| `ResolveSymlinkDest` returns some other error | ❌ unpinned |
| refuse the chmod even with `AGENTSYNC_ALLOW_SYMLINK_DEST=1` | ❌ unpinned |
| drop the keep-set `continue` in `applyPlan` | ✅ caught |
| drop the `Writer.Write` / `ReadFileOptional` shape gates | ✅ caught |

the first row is the one I'd most like covered, since it's the data-loss case this PR actually fixes: one small table test per adapter that makes the merge target unreadable and asserts an error rather than an overwrite.

smaller test things:
- `TestWrite_ChmodRefusesSymlinkDest`: assert `errors.Is(err, iox.ErrSymlinkDest)` and add the opt-in positive case. `fi, _ := os.Stat(target)` will also nil-panic if the stat ever fails.
- `TestWriterWrite_FIFODoesNotBlock`: check the FIFO still exists afterward. it also landed directly under `TestWriterDelete_FIFODoesNotBlock`'s doc comment, so it took that comment over and `TestWriterDelete` is now undocumented.
- `readoptional_unix_test.go` needs `testenv.RequireContainer(t)` like its siblings. right now it stays off the host only because its name happens to match the `-skip 'TestReadFileOptional'` in the justfile.
- `FilterOrphanDeletes` in `baselinePaths` is a no-op (kept paths are already in `out` from `res.Ops`), so drop it or comment why it's there.
- no case where the dropper sorts *before* the keeper. there it's delete-then-rewrite, which passes with or without the fix.

### Not blocking, fine as follow-ups

- `pipeline.go:480` reimplements `FilterOrphanDeletes` inline, right next to a comment saying both must use the same rule. one exported helper that both apply and the CLI call would remove the drift risk, and it would keep the HOME-relative `map[string]struct{}` out of the public API.
- "not a regular file" now has three spellings: `adapter.ErrNotRegularFile`, `cli.errDestNotRegular`, and a bare string at `writer.go:204`. at minimum, wrap the sentinel there; long term I'd probably put one sentinel in `iox` next to `ErrSymlinkDest`.
- dry-run marks a mode-only change on a symlinked dest as would-change, but the real apply refuses it. the content-differs path already behaves this way, so it isn't new.
- `FilterOrphanDeletes` reuses its input's backing array (`dels[:0]`), which is surprising in an exported function.
- nit: the "Shape first" comment at `writer.go:201` is spliced into the middle of the existing convergence comment.

### What I'd like to do

keep it as one PR is fine. fix the two _blockers_ and _issues_ the adapter read-error tests are the other thing I'd want before merge; everything under "not blocking" can come later.

really appreciate you picking up three of the gnarlier open issues in one go 🚀. ping me once it's updated and I'll re-review!

harryy2510:
@spxrogers thanks for this one too — both blockers and all four issues are addressed, pushed as `36248e3` + `86acc4e`, and both hermetic legs are green.

### BLOCKER 1 — the entry that never retired

Reproduced exactly as you described. Keeping the file was necessary and not sufficient: the dropping agent's state entry is held by `PruneStaleState`'s reclaimable arm, which `Lstat`s the path — and the keeper had just rewritten it — so the entry survived every run and `status` re-offered the orphan forever.

Rather than patch the two sites, I gave the plan-derived keep-set a name: `render.SharedDests`. Four consumers now share it instead of each deriving its own answer —

- `applyPlan` skips the delete,
- `PruneStaleState` **releases** the dropping agent's entry (the delete that would have retired it is never going to run),
- `OrphanFiles` omits the path, so `status` / `diff` stop promising a deletion,
- `removalCounts` agrees with all three.

That also absorbs your "not blocking" note about `pipeline.go` reimplementing `FilterOrphanDeletes` inline next to a comment saying both must use the same rule — there is one implementation now. `FilterDeletes` returns a fresh slice too; the `dels[:0]` aliasing is gone and pinned.

`baselinePaths` lost the filter entirely rather than gaining a comment. You were right that it is a no-op — a kept delete is kept *because* some agent still renders that path, so the ops loop above already added it — and I would rather delete a line that reads like a safety check than explain why it is not one.

**This changed two characterization tests, one named `orphans-are-not-cross-agent-filtered`.** Its comment said the exclusion "is reconcile's, not the walk's". I read that as the bug rather than the contract — reconcile excluded it, apply kept the file, and only status/diff were left promising a deletion nobody would perform. Both tests now assert the exclusion and say why the rule moved. Flagging it explicitly since I overturned a documented decision; if that split was deliberate for a reason not in the comment, this is the commit to push back on.

### BLOCKER 2 — docs

All of it: `SECURITY.md` and `architecture.md` no longer list #248 as an open gap, `components.md` no longer says `Writer.Write` and the ingest paths still block, and the CHANGELOG's "are NOT fixed by this" paragraph now points at new `Fixed` entries for #246, #241/#242 and #248, plus one for the `readJSONFile` data-loss fix, which was user-visible and had no entry at all. Stale comments in `destread.go`, `reconcile.go`'s `destReadRefusal` (the advice is unchanged but the *reason* is now "it refuses" rather than "it hangs"), and the four `readJSONFile` doc comments.

### #242 — finished the sweep

Took the sweep over "Refs #242". Every adapter `Ingest` now reads through `adapter.ReadFileOptional`: claude, codex, cursor, gemini, opencode, roo, windsurf, cline, continuedev, generic, and `codex.IngestPlugins`. `import claude` with a FIFO at `~/.claude/agents/x.md` and `import codex` with one at `~/.codex/prompts/x.md` both return now.

The `--auto-override` skip is gone and the row passes, and I added the `doctor` row.

### Tests

Your top ask first: `readJSONFile` surfacing a read error is pinned per adapter across claude, cline, cursor, roo and windsurf, each with an absent control and an unparseable control so the parse-still-degrades-to-empty half stays honest.

Also `TestWrite_ChmodRefusesSymlinkDest` matching `iox.ErrSymlinkDest` rather than any error (with the nil-panic `Stat` fixed) plus the opted-in positive row; `TestWriterWrite_FIFODoesNotBlock` asserting the FIFO survives and getting its own doc comment back; and `readoptional_unix_test.go` calling `testenv.RequireContainer` instead of relying on its name matching the justfile's `-skip`, which was the part of that finding I found most alarming.

### Still open, deliberately

- The `applyWrite` → `os.ReadFile` mutation for **codex / gemini / generic / opencode**. They already go through `adapter.ReadExisting`, but they have no `readJSONFile` helper to aim a test at the way the other five do, so those four rows stay unpinned.
- `FilterOrphanDeletes` dropped from `removalCounts`, and `ResolveSymlinkDest` returning some other error.
- The dropper-sorts-before-keeper ordering case.
- Your three remaining follow-ups: the three spellings of "not a regular file", dry-run's mode-only-on-symlink preview, and the spliced "Shape first" comment.

Say which of those you want in this PR and I will add them; otherwise I will open a follow-up.

PR #271, opened 2026-09-19
fix(testenv): hermetic suite under a configured shell — AGENTSYNC_TARGET_ROOT sandbox, ambient-env scrub, GROK_HOME guard (#270)

### LINKED ISSUE #270: test suite is not hermetic: an ambient AGENTSYNC_HOME reds out 535 tests, including the secret-leak guard
## Summary

`go test ./...` passes only in an environment where none of our variables are set. CI is exactly that environment, so the suite has never been validated against a machine that actually *uses* agentsync — and someone dogfooding the tool is the single most likely person to have `AGENTSYNC_HOME` exported.

Measured on `1cc3b9a`, same container, only the environment differs:

| ambient env | failing tests on main |
|---|--:|
| *(clean)* | 0 |
| `NO_COLOR=1` | 2 |
| `AGENTSYNC_HOME=<path>` | **535** |

All 535 are in a single package: **`internal/cli`**. They include `TestCapture_ImportNoSecretLeak` — the guard that exists to prove a resolved secret never reaches canonical source. It fails for an environment reason rather than a real leak, which is the worst possible failure mode for that particular test: it trains us to ignore it.

**This is a test-isolation leak, not a defect in the shipped CLI.** No user is affected. See root cause.

## How this was found — credit where it's due

This surfaced out of #261, @harryy2510's Grok Build adapter. Reviewing that PR turned up a single failing guard (`TestEveryAdapterRejectsTraversalComponentName/grok/legitimate/user`) whenever `GROK_HOME` was exported. Pulling that thread showed grok wasn't the problem at all — it was the third instance of a pre-existing class, and by far the smallest:

| ambient env | main (`1cc3b9a`) | with #261 |
|---|--:|--:|
| *(clean)* | 0 | 0 |
| `NO_COLOR=1` | 2 | 2 |
| `GROK_HOME=<path>` | — | 1 |
| `AGENTSYNC_HOME=<path>` | **535** | 542 |

#261 contributes one additional failing test to a pile that already stood at 535. Nice find, @harryy2510 — a genuinely useful thing to shake out of a first contribution. 🙏

## Repro

```
export AGENTSYNC_HOME=/tmp/some-dir
AGENTSYNC_TEST_IN_CONTAINER=1 go test ./...
# => 535 failures, all in internal/cli
```

Unset it and the suite is green.

## Root cause

A precedence inversion between the product contract and the test-isolation contract, in `internal/paths/paths.go`:

```go
func HomeDir(e Env) string {
    if root := e.Get("AGENTSYNC_TARGET_ROOT"); root != "" { return root }  // tests redirect here
    return e.Get("HOME")
}

func AgentsyncHome(e Env) string {
    if h := e.Get("AGENTSYNC_HOME"); h != "" { return h }                   // ...but this wins outright
    return filepath.Join(HomeDir(e), ".agentsync")
}
```

`AGENTSYNC_HOME` beating the home fallback is **correct product behaviour** — an explicit override should win. The bug is that tests redirect via `AGENTSYNC_TARGET_ROOT`, which `AgentsyncHome` never consults, so an ambient `AGENTSYNC_HOME` silently outranks the redirect and every test resolving a canonical source lands on the developer's real tree instead of its temp dir.

## Why CI never sees it

CI (`.github/workflows/ci.yml`) runs `just test-fast`, `just test-release` and `just lint` in a pristine runner/container. There is no leg that runs the suite the way a configured machine looks, so none of this is visible until someone runs the tests on a box that uses the tool.

## The sibling instance: `GROK_HOME`

`internal/cli/registry_internal.go:53` reads `os.Getenv("GROK_HOME")` raw. Worth noting that `registry_internal.go:35` already resolves `home` via `paths.HomeDir(paths.OSEnv{})` — the injectable abstraction is one line above. Every other ambient `os.Getenv` in production code is either an `AGENTSYNC_*` escape hatch, a universal convention (`NO_COLOR`, `EDITOR`), or routed through `paths.Env`; `GROK_HOME` is the only third-party variable read raw.

Separately, `GROK_HOME` is validated only by `filepath.IsAbs` (`internal/adapter/grok/ingest.go:132`), so `/` or `$HOME` is accepted and flows into `VersionRoots` → `enabledVersionRoots` de-nesting (which collapses every other agent's root into an ancestor) → `git init`. Measured: `GROK_HOME=/` yields `VersionRoots=[/]`; `GROK_HOME=$HOME` collapses 26 roots into 1.

Two options, and this issue should settle which:
- **(a)** bound it — reject `/`, reject `== $HOME`, require a component below the root. Costs a legitimate upstream feature, since pointing `GROK_HOME` outside `$HOME` is exactly what upstream intends it for.
- **(b)** accept that an env-derived root legitimately escapes `$HOME`, and teach `TestEveryAdapterRejectsTraversalComponentName` about it — combined with (2) below so the suite stays deterministic.

(b) is the leaning, but it is a real design call, not a mechanical fix.

## Proposed fix

1. **Fix the precedence** so a test's `AGENTSYNC_TARGET_ROOT` redirection outranks an ambient `AGENTSYNC_HOME`, without changing the CLI's user-facing behaviour (explicit override must still win for real users). Likely accounts for the whole 535 — **confirm by applying and re-running, do not assume.**
2. **Neutralize ambient env in the harness.** `internal/testenv` is the natural home — its own package doc already says it "contains helpers used by tests to assert hermeticity preconditions", and `RequireContainer` is called from the FS-touching packages. Clearing/overriding the set of variables that can move a result belongs right there.
3. **Route `GROK_HOME` through `paths.Env`** as part of (2), so it is injectable like everything else.
4. **Add a CI leg with config fidelity** — one matrix entry with `AGENTSYNC_HOME`, `GROK_HOME` and `NO_COLOR` set would have caught all three of these and would keep catching them.

Also in scope: `internal/cli/gitbackup.go:367` asserts *"every version root is a hardcoded string literal with fixed casing ... if a future adapter derives a root from a case-varying source, this key would need case-folding"*. Grok is now that adapter, so either the sentence changes or the dedup gains case-folding.

## Definition of done

- `AGENTSYNC_TEST_IN_CONTAINER=1 go test ./...` is green with each of `AGENTSYNC_HOME`, `GROK_HOME`, `NO_COLOR` set, individually and together.
- Still green with none set.
- The CLI's own `AGENTSYNC_HOME` override behaviour is unchanged for real users, with a test pinning it.
- A CI leg exercises the configured-environment case.

## Conventions for whoever picks this up

Read `CLAUDE.md` first. The load-bearing ones here:

- **Stdlib testing only** — no testify/gomega. Table-driven with a `name` field and `t.Run`; `t.Helper()` in helpers that call `t.Fatal/Errorf`.
- **Filesystem in tests** is `afero.NewMemMapFs()` or `t.TempDir()`, never `os.UserHomeDir()` (a `forbidigo` rule bans it in `_test.go`; use `paths.HomeDir(env)`). FS-touching tests call `testenv.RequireContainer(t)`.
- **Docs are part of the contract.** If this changes the `~/.agentsync/` layout story or the env-var semantics, update `docs/concepts.md` / `docs/architecture.md` §2 and the layout block in `docs/user-guide.md` in the *same* commit, plus `CHANGELOG.md` under `[Unreleased]` for anything user-visible.
- **`just lint` rewrites files in place** (`gofmt -s`, gofumpt, `go mod tidy`) — re-stage anything it touches before committing, or CI's `git diff --exit-code` fails. In a container session the toolchain must be forced: `GOTOOLCHAIN=go1.26.2 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run ./...`.
- Conventional commits with scope, e.g. `fix(testenv): ...`.

## What not to do

- Don't "fix" the 535 by weakening or skipping `TestCapture_ImportNoSecretLeak` or any other guard — they are failing for an environment reason and must stay load-bearing.
- Don't change `AgentsyncHome`'s precedence in a way that breaks the documented user-facing override.
- Don't fold this into #261. That PR is merging on its own merits; its `GROK_HOME` instance is 1 of 536 and belongs here.

spxrogers:
## Summary

Closes #270.

`go test ./...` passed only in an environment where none of agentsync's own variables were set — which is exactly what CI is — so the suite had never been run the way a machine that actually *uses* agentsync runs it. Reproduced on this branch's base: an exported `AGENTSYNC_HOME` failed 465 `internal/cli` tests (535 on the issue's machine), `TestCapture_ImportNoSecretLeak` among them; `NO_COLOR=1` failed two `internal/ui` colour tests; an exported `GROK_HOME` moved the grok adapter's root out from under `TestEveryAdapterRejectsTraversalComponentName`.

Root cause, as the issue diagnosed: `paths.AgentsyncHome` never consulted `AGENTSYNC_TARGET_ROOT`, so an ambient `AGENTSYNC_HOME` outranked the tests' redirect and every canonical-source lookup landed on the developer's real `~/.agentsync`. Five tests had already been hand-patched with `"AGENTSYNC_HOME": ""` to dodge it.

### The fix, per the issue's four points

1. **Precedence.** `AGENTSYNC_TARGET_ROOT` is a real sandbox: while set, the canonical source resolves to `<root>/.agentsync` and `AGENTSYNC_HOME` is ignored. Real users never set the redirect, so their `AGENTSYNC_HOME` override is unchanged (`TestAgentsyncHome_Precedence` drives a real `init` both ways). New `paths.AgentHomeOverride` applies the same sandbox rule to third-party agent home variables.
2. **Harness scrub.** Importing `internal/testenv` scrubs, at package init, every `AGENTSYNC_*` override (except the harness's own `AGENTSYNC_TEST_*` / `AGENTSYNC_LIVE_*`) plus `GROK_HOME`, `NO_COLOR`, `EDITOR` — before any `TestMain` or `t.Setenv`. `internal/ui` imports it for the side effect. `TestScrubRunsAtInit` re-execs the test binary to pin the wiring in a pristine run. `internal/cli`'s `TestMain` also runs under a throwaway `HOME` and fails the package if any test writes there (one did).
3. **`GROK_HOME` through `paths.Env`.** The registry reads it via `paths.AgentHomeOverride`; a go/ast guard (`TestAgentHomeVarsReadOnlyThroughPaths`) rejects a raw `os.Getenv` of an agent-home variable, and `TestEnvReadingPackagesScrubAmbient` requires every env-reading production package's tests to import `testenv`.
4. **Configured-environment CI legs.** `test-release` runs twice (`just test-release` / `just test-release-configured`), `test-fast` gains one configured Ubuntu entry. The configured container leg exports `AGENTSYNC_HOME`/`GROK_HOME`/`NO_COLOR`/`EDITOR` and puts an exit-0 stub for all 28 agent binaries agentsync probes on `PATH` (the one ambient input the scrub cannot cover), self-checks at runtime, and exports a marker that `TestConfiguredLegIsLive` observes from inside a test. Parity guards keep the variable list, the stub list, and the recipe wiring in step with their sources of truth, scoped to the shell/YAML blocks they belong in.

### The (a)-vs-(b) design call: **b, plus a strong guard**

A `GROK_HOME` outside `$HOME` is what upstream provides the variable for, so it stays legitimate and versioned (b). But the hazard the issue measured is real — `GROK_HOME=$HOME` collapsed 26 version roots into one and would `git init` the home directory — so two layers close it:

- **Grok refuses `GROK_HOME=/` and `GROK_HOME=$HOME`** on every command (`validateHome`, the single gate behind `Detect`/`Render`/`Ingest`/hooks/`VersionRoots`), with an error naming the variable and suggesting `~/.grok`. `$HOME` is compared by directory identity (`paths.SameDirResolved`), so a symlinked or, on macOS/Windows, case-varied spelling is refused too.
- **The apply tail never inits at or above `$HOME`, for any adapter** (`partitionVersionRoots`): a declared root that contains the user's home by identity (`paths.ContainsDirResolved`) *or* by spelling (`paths.ContainsDir`, so `/home` is refused even when `$HOME=/home/alice` is itself a symlink to `/data/alice`) is dropped before de-nesting, with a warning; `doctor` reports it and `revert` skips it. De-nesting itself stays lexical by declared path — resolving symlinks there was tried during review and measured to break `revert` (see the review-loop summary comment).

### Docs

README + website env tables (`AGENTSYNC_TARGET_ROOT` sandbox row, `GROK_HOME` row), `docs/architecture.md` §7, `docs/components.md` (`paths`, `testenv`), `docs/grok.md`, `CHANGELOG.md` (two `[Unreleased]` Fixed entries), and the canonical memory `.agentsync/memory/AGENTS.md` with `CLAUDE.md`/`AGENTS.md` re-rendered via `agentsync apply --scope project`.

### Review

Eight rounds of the repo's `review-loop` skill (four Opus lenses per round); see the summary comment for the round-by-round findings. Every accepted finding was break-verified before it was fixed.

## Type of change

- [x] Bug fix
- [ ] New feature / enhancement
- [ ] Refactor (no behavior change)
- [x] Docs
- [x] Tests / CI / tooling

## Test plan

Verified in the session container (`AGENTSYNC_TEST_IN_CONTAINER=1`; neither podman nor docker was available, so this PR's CI is the first execution of the container gate's two legs):

- `go test ./...` green with **none** of `AGENTSYNC_HOME` / `GROK_HOME` / `NO_COLOR` / `EDITOR` set.
- `go test ./...`, `-tags=e2e`, `-tags=bdd` green as the **configured leg**: all four variables exported, 28 fake agent binaries first on `PATH`, `AGENTSYNC_TEST_AMBIENT_BIN` set (so `TestConfiguredLegIsLive` ran).
- Host `test-fast` package set green with the variables exported, `-race`.
- `-race` on the touched packages (`paths`, `testenv`, `grok`) in both legs.
- Lint recipe (`gofmt -s`, gofumpt, `go mod tidy`, golangci-lint v2.12.2 under `GOTOOLCHAIN=go1.26.2`): 0 issues, nothing rewritten.
- Every guard was mutation-tested in a throwaway clone during review (listed per round in the summary comment).

- [ ] `just test-release` is green (the release bar) — runs in CI on this PR, both legs
- [x] `just lint` is clean

## Checklist

- [x] Conventional commit messages with a scope (e.g. `fix(secrets): …`).
- [x] Tests added/updated for the behavior changed.
- [x] If this touches `internal/secrets`, `internal/capture`, or any `source.Write*` path, I've re-read the secret-handling invariants in `CLAUDE.md` / `SECURITY.md` and not weakened them. (Not touched; `TestCapture_ImportNoSecretLeak` stays load-bearing and now runs against the tmp tree it was always meant to.)
- [x] Docs updated if behavior, CLI surface, or capability coverage changed.

https://claude.ai/code/session_01CTYGFky2yQbr7evC3FPKby


spxrogers:
## Review-loop summary (8 rounds, four Opus lenses per round)

This PR went through the repo's `review-loop` skill: correctness, adversarial, API-design and test-rigor reviewers run read-only in parallel each round; every accepted finding was fixed in the main session, break-verified, and committed as its own round. Budget was 10 rounds; the loop stopped at 8 on the first NIT-only round with all four lenses at `CLEAN — ship it`.

| Round | Commit | Worst finding | What changed |
|---|---|---|---|
| 1 | `08c72aa` | ISSUE | Scrub moved to `testenv` package `init()` (a per-test scrub could wipe a redirect a test had just set); central never-at-or-above-`$HOME` guard in `partitionVersionRoots` with warnings in apply/doctor and a filter in revert; one containment predicate; a cli test that wrote into the real `$HOME` fixed and the package's `TestMain` now runs under a throwaway `HOME` and fails on strays; three vacuous grok rows made real; ambient-var parity guards. |
| 2 | `4cbd826` | ISSUE (3 lenses) | Round 1's symlink resolution resolved only fully-existing paths, so a first-apply child under a symlinked parent was no longer de-nested. Fixed by resolving through the deepest existing ancestor. |
| 3 | `0ab55e8` | ISSUE (3 lenses) | Round 2's "symlinked-out child gets its own repo" was measured false (Detect walks the link spelling into the parent's `.git`) and made the parent un-revertable. Reverted: de-nesting is lexical (`paths.ContainsDir`); identity resolution (`ContainsDirResolved`/`SameDirResolved`) only backs the `$HOME` guard and the `GROK_HOME` refusal. |
| 4 | `d893887` | ISSUE | With `$HOME` itself a symlink (`/home/alice → /data/alice`), `GROK_HOME=/home` passed the identity-only guard. The guard now drops on spelling OR identity. A relative-path slice bug in the resolver fixed (latent). |
| 5 | `ede155b` | ISSUE | `PATH` is the one un-scrubbable ambient input: a cli test asserted the "codex not found on PATH" warning and passed only because CI has no codex. Test neutralizes `PATH`; the configured container leg fakes all 28 probed agent binaries; a parity guard keeps that list in step. |
| 6 | `b589bc0` | **BLOCKER** | Round 5's leg was inert: the runner never forwarded `AGENTSYNC_TEST_CONFIGURED_ENV` into the container, and a text-only guard could not see it. Forwarded, plus a runtime self-check in the entrypoint. |
| 7 | `e225886` | ISSUE | Round 6's guard was a whole-file substring check; forwarding the flag unconditionally (pristine leg configured too) stayed green. Guards scoped to the shell if-blocks; the entrypoint exports `AGENTSYNC_TEST_AMBIENT_BIN` and `TestConfiguredLegIsLive` observes from inside a test that `exec.LookPath` resolves the stubs. |
| 8 | `2aa76e3` | NIT only | Header match anchored to line start (one measured false-green path closed); ci.yml check scoped to the configured step; comments and messages. |

**Where the risk was.** Two of the seven defect-bearing rounds (3 and 6) were defects introduced by the previous round's own fix, both caught by the next round measuring rather than reading. The original #270 fix (redirect precedence, scrub, `GROK_HOME` routing) never changed after round 1; the churn was in the review-added `$HOME` guard's containment semantics and in the CI harness plumbing, which is now pinned behaviourally rather than by text where a runtime fact exists.

**Verified at `2aa76e3`:** full suite green pristine; green as the configured leg (ambient vars, 28 stubs on `PATH`, marker set); e2e and bdd green under the stub `PATH`; race clean on touched packages; lint 0. The container gate itself (`just test-release` / `test-release-configured`) has not run here — no podman/docker in this session — so this PR's CI is its first execution of both legs.

**Follow-ups noted by reviewers, deliberately not in this PR:** the website env page's "every `AGENTSYNC_*` override" sentence is already false on `main` (`AGENTSYNC_LOCK_TIMEOUT_MS`, `AGENTSYNC_TEST_DEBUG` unlisted); README cites a `TestApply_FirstRun` that does not exist on `main`.

---

PR #261, opened 2026-09-10
feat(adapter): add Grok Build support

harryy2510:
## Summary

Add a dedicated `internal/adapter/grok` adapter for Grok Build, registered as `grok` with directory/binary detection and `GROK_HOME` support.

- Support user and project instructions/AGENTS.md, complete skill directories, Markdown commands, TOML MCP, and JSON command hooks.
- Reuse existing merge, skill, and capture code; keep hook parsing inside Grok without modifying other adapters. Add an optional path-specific merge strategy so cleanup/purge handles Grok's TOML and JSON files correctly. Preserve Grok headers during MCP reconciliation.
- Update docs, website coverage, and compatibility matrices with verified upstream references.

Intentional limits: hook import manages only `hooks/agentsync.json` to avoid duplicate execution; unsupported events, HTTP handlers, and unmodeled handler fields are reported/refused. Project trust remains Grok-owned. SSE transport labels normalize to HTTP; TOML comments are not preserved. Subagent/LSP projection and native plugin discovery are deferred. Full details: [Grok support](https://github.com/harryy2510/agentsync/blob/feat/grok-build-adapter/docs/grok.md).

## Type of change

- [x] New feature / enhancement
- [x] Docs
- [x] Tests

## Test plan

- [x] Existing release-test container entrypoint: vet, build, race tests, lifecycle e2e, BDD, and CLI smoke checks.
- [x] Pinned golangci-lint (0 issues), scoped formatting, and read-only `go mod tidy -diff`.
- [x] Native-file round trips, both-scope ownership/cleanup/purge, MCP import/reconciliation with secret re-referencing, and enriched-hook retirement.
- [x] Website documentation generation, link checker (0 broken; existing allowlist retained), and 7 script tests.
- Live Grok sessions were not exercised.

## Checklist

- [x] Conventional commit with scope.
- [x] Tests and capability documentation updated.
- [x] Existing secret/capture boundary preserved; no canonical schema or dependency changes.

Issue #124, opened 2026-06-30
fix(claude): preserve unmodeled hook fields and stop corrupting non-command handlers on import→apply


spxrogers:
> [!IMPORTANT]
> **Update since this issue was filed — PRs #73 and #180 merged after the epic (#178) was created.**
> **PR #73** ("fix: skip unsupported Claude LSP settings", which resolved the pre-existing #66) **removed Claude's LSP support entirely**: it deleted `internal/adapter/claude/lsp.go`, dropped `CapLSP` from Claude's `Capabilities()` bitmask, stopped writing/reading `settings.json#/lspServers`, and now emits an `adapter.Skip` for Claude LSP — with `capability-matrix.md`, `architecture.md`, `components.md`, `user-guide.md` and the tests updated in the same commit. **PR #180** re-homed the render-pipeline sibling-section / orphan-cleanup regression tests onto Codex and touched `internal/render/pipeline.go`.
>
> Because these commits changed `claude/{render,ingest,claude,paths,render_test,ingest_test}.go`, several `docs/*.md`, and `render/pipeline.go`, **the exact `file:line` references below have drifted — locate code by symbol, not by line number.**
>
> **Impact on THIS issue:**
> - **The hook-model bug is unchanged — this is still a blocker.** But #73 changed `claude/render.go` (Claude now writes a **single** key-merge section to `settings.json`) and `claude/ingest.go` (−26 lines; LSP read removed), and reworked `ingest_test.go`. Re-locate `renderHooks`/hook-ingest by symbol; the `TestIngest_RoundTripsHooksAndLSP` reference (in "Acceptance criteria" and "Why no test caught it") now points to a renamed/hooks-only test — verify its current name. The `LSPServerSpec.Extra` analogy for the `Extra`-passthrough option still holds (the canonical schema field remains).

---

## Metadata

- **Severity:** blocker
- **Area:** adapters
- **Wave:** 1
- **Themes:** C1 (artifact-fidelity / lossy-conversion class)
- **Depends on:** none
- **Can run in parallel with:** all other Wave-1 work. Touches only `internal/adapter/claude/{hook.go,ingest.go}` + the `source.Hook` struct in `internal/source/schema.go`. **Gates** #147 (capability-matrix Hook-cell downgrade) and the Claude hook-fidelity sweep in #153 — leave those for their own issues.
- **Labels:** `severity:blocker`, `area:adapters`, `theme:C1`, `wave:1`

## Summary

[...]

PR #273, opened 2026-09-20
feat(hooks): model command-handler timeout across adapters

harryy2510:
## Summary

Models a command hook handler's **timeout** on the canonical `source.Hook`, so a
native `timeout` is rendered, ingested, and round-tripped instead of being treated
as an unmodeled field that forces agentsync to refuse the whole event.

Before this change, `timeout` was in the "reported, not silently dropped" bucket:
a native Claude/Cursor/Gemini hook carrying one was left uncaptured with a warning
on import, which meant a user who set a timeout natively could not bring that hook
under agentsync management at all.

- `source.Hook` gains `Timeout int` (seconds; `0` omits the native field), with
  loader + writer support so `hooks/<event>.toml` persists it losslessly.
- Two shared helpers in `internal/adapter/keys.go` keep the parse/emit rules in one
  place: `ParseHookTimeout` (absent is fine; a present negative, fractional, or
  non-numeric value is structural/malformed) and `SetHookTimeout` (emits only when
  `> 0`, so existing canonical sources render byte-identically).
- Render + ingest wired through **Claude, Codex, Cursor, Gemini, and Grok**.
- `codex.MergeTOML` now converts `json.Number` to `int64`/`float64`. Without this,
  a numeric value decoded from JSON serialized into `config.toml` as a *string*,
  so the Codex hook timeout would have rendered as `timeout = "25"`.
- `import`'s `hookSignature` includes the timeout, so changing only the timeout
  natively is correctly seen as a change rather than a no-op.

## Type of change

- [ ] Bug fix
- [x] New feature / enhancement
- [ ] Refactor (no behavior change)
- [ ] Docs
- [ ] Tests / CI / tooling

## Test plan

Round-trip coverage added per adapter (render → on-disk native shape → ingest →
canonical), plus source-level persistence and the shared parser:

- `internal/source/writer_test.go` — timeout survives the canonical TOML round trip.
- `internal/adapter/keys_test.go` — `ParseHookTimeout` across `int`, `int64`,
  `float64`, `json.Number`, absent, negative, fractional, and non-numeric.
- `internal/adapter/{claude,codex,cursor,gemini,grok}` — per-adapter round trips,
  including `TestHookTimeoutRoundTrip` asserting Grok's on-disk JSON carries a
  numeric `timeout`.
- `internal/adapter/codex/settings_test.go` — `MergeTOML` keeps numbers numeric.
- `internal/cli/import_hook_signature_internal_test.go` — a timeout-only edit
  registers in the signature.
- Guard tests in each adapter updated so an unmodeled-field refusal no longer
  counts `timeout` among the unmodeled keys.

Verification run:

- [ ] `just test-release` is green (the release bar) — **not run locally**: no
      container runtime was reachable on this machine, so the filesystem-gated
      packages did not execute here. Relying on CI for that leg.
- [x] `just lint` is clean — `gofmt -s`, `gofumpt`, and `golangci-lint v2.12.2`
      all report no issues.
- [x] `go build ./...` and `go vet ./...` pass.
- [x] Host-safe unit set (`just test-fast` package list) passes with `-race`.

## Checklist

- [x] Conventional commit messages with a scope (e.g. `fix(secrets): …`).
- [x] Tests added/updated for the behavior changed.
- [x] If this touches `internal/secrets`, `internal/capture`, or any
      `source.Write*` path, I've re-read the secret-handling invariants in
      `CLAUDE.md` / `SECURITY.md` and not weakened them.
      `Hook.Timeout` is an `int`, not a string-shaped field, so it is not
      secret-bearing and `walkSecretFields` is deliberately unchanged;
      `TestNewSecretFieldGuard` still passes.
- [x] Docs updated if behavior, CLI surface, or capability coverage changed.
      `docs/architecture.md`, `docs/capability-matrix.md`, and `docs/grok.md` no
      longer cite `timeout` as an unmodeled field; `CHANGELOG.md` updated under
      `[Unreleased]`.


spxrogers:
heya @harryy2510, thanks for splitting this out so fast after #261 🙏 I ran it through the same four-lens [review-loop](https://github.com/spxrogers/agentsync/blob/main/.agentsync/skills/review-loop/SKILL.md) plus my own pass.

short version: **this closes every item I asked for on #261**: the `MergeTOML` fix, loader/writer round-trip, render coverage for codex/cursor/gemini, and `hookSignature`. CI is green and it merges clean against `main`. but there's one blocker in how the value maps onto Gemini, plus a few "never drop silently" gaps I'd like closed before it lands.

### What's strong

- **the `MergeTOML` fix is in the right place.** it also closes the pre-existing codex MCP extras bug (`startup_timeout_sec = '10'` → `10`), which is exactly what I was hoping for.
- **`Timeout = 0` renders byte-identical to `main`**, so nobody's existing canonical source drifts on upgrade.
- **the secret boundary is intact.** `int` isn't string-shaped, so no `walkSecretFields` entry is needed, and `TestNewSecretFieldGuard` is telling the truth.
- **the wiring is well pinned.** deleting `SetHookTimeout`, the ingest assignment, the modeled-keys entry, `ConvertNumbers`, the signature field, or the loader/writer field each fails a test, in every adapter. nice.

### BLOCKER: Gemini's `timeout` is milliseconds, not seconds

from the [Gemini CLI hooks reference](https://geminicli.com/docs/hooks/reference/): *"Execution timeout in milliseconds (default: 60000)."* Claude, Codex, Cursor and Grok are all seconds (I re-checked each upstream). `gemini/hook.go:128` (render) and `gemini/hook.go:304` (ingest) pass the canonical value straight through:

| direction | in | out | effect |
|---|---|---|---|
| apply | canonical `timeout = 30` (s) | gemini `"timeout": 30` | 30**ms**, hook gets killed |
| import | gemini `"timeout": 60000` (its default) | canonical `60000` (s) | fans out to claude/codex/cursor/grok as ~16.7h |

the fixture this PR swapped out in `gemini/apply_test.go:176` was literally `"timeout": 5000`, which is ms-scale. `TestRoundTrip_Hooks` only goes gemini → gemini, so the mismatch cancels itself out and stays green.

fix: ×1000 on render, ÷1000 on ingest. a native value that isn't a whole second (e.g. `1500`) should be refused rather than rounded. please pin it with an on-disk assertion (`"timeout": 45000`) plus one cross-adapter test (e.g. claude native → gemini render). that's the test that would have caught this.

### ISSUE: numbers we can't represent are classified as typos

`ParseHookTimeout` (`keys.go:42`) returns `structural = true` for *every* failure, including valid JSON numbers like `1.5`, `30.0`, `1e2`, and negatives. structural means "malformed, never retire". so an event that was captured clean and later gets `timeout: 1.5` natively no longer triggers stale-hook retirement, and the next apply rewrites the user's native array without it. that's the #124 hole the semantic/structural split exists to close. before this PR, *any* `timeout` was a semantic refusal, so this path actually regresses.

only a type mismatch (string / bool / object / null) should be structural. a number we can't represent should be a semantic refusal, like any unmodeled field. once that's split, the third return is also redundant as written (it's always `!ok`), so `(int, bool)` or a small enum reads better. small nit while you're in there: the warning says "not an integer" for negatives too.

### ISSUE: three silent drops (CLAUDE.md "capture it or acknowledge it")

- **explicit `timeout: 0`** ingests as `0`, then re-renders with the key removed, so it silently becomes the harness default. either refuse a present `0` or model it as `*int`. either way, write the choice down.
- **a negative canonical timeout** (`timeout = -5` in `hooks/<event>.toml`) loads fine, and `SetHookTimeout` quietly omits it. reject it in the loader.
- **plugin projection**: `appendHookGroup` / `appendCommandHook` (`marketplace/projection.go:~1036`) never read `timeout`, so a plugin's hook timeout is dropped now that the model can carry it. `dedupHooks`' key (`projection.go:247`) needs it too once it's threaded through. `hookSignature` matching only works today because both sides are always `0`.

### Docs / comments

- stale "timeout is unmodeled" comments:
  - `claude/hook.go:84`, plus the "APPROACH A (no schema change)" block right below it. there is a schema change now 🙃
  - `claude/ingest.go:19`
  - `cursor/hook.go:129`
  - `gemini/hook.go:168,185`
  - `grok/hook_ingest.go:19`
  - `hookSignature`'s doc comment (`import.go:1242`) still describes four fields.
- `docs/components.md:243` still uses `timeout` as its unmodeled-field example, and `capability-matrix.md:259` still says hooks model "`matcher` + `command`".
- this is a canonical schema change, so per the CLAUDE.md table it needs `docs/concepts.md`, `architecture.md` §2, and the layout block in `user-guide.md`. none of them mention the new `timeout` key, or that it's **seconds**, which matters more now that one harness isn't.
- CHANGELOG: the `MergeTOML` number fix is user-visible on its own (codex MCP extras), so it deserves a `### Fixed` entry instead of being folded into Added. the Added bullet also reads a bit like a commit log; tell users what they get.
- the PR description says `keys_test.go` covers `int64`, `float64` and fractional values, but `TestParseHookTimeout` only has missing / int / json.Number / negative / string.

### Tests

- `TestParseHookTimeout` (`keys_test.go:47`): the `json.Number` branch is the one real decoding hits (`UseNumber` everywhere), but only its happy path is tested. removing its `i < 0` check, the `int64`/`float64` negative checks, or the fractional check breaks nothing. please make it table-driven with a `name` field and cover:
  - `json.Number`: `"-1"`, `"25.0"`, `"1e2"`, `"99999999999999999999"`
  - `int64`
  - `float64`: `30.5`, `-3`
  - `bool` / `nil`
- codex is the only adapter without a malformed-timeout row in `hook_guard_test.go`.
- `writer_test.go` round-trips the model. please also assert the TOML bytes: `timeout = 10` present, and the key absent at `0`.
- small: chained bare type assertions (`claude/hook_test.go:382`, `grok/grok_test.go:564`, `codex/settings_test.go:184`) panic instead of failing cleanly.

### To land

- [ ] Gemini ms ↔ s conversion + on-disk and cross-adapter tests
- [ ] structural vs semantic split in `ParseHookTimeout`
- [ ] the three silent drops (explicit `0`, negative canonical, plugin projection)
- [ ] stale comments, contract/schema docs, CHANGELOG `Fixed` entry
- [ ] table-driven `ParseHookTimeout` test + codex malformed-timeout row

fine as follow-ups:
- a CLI-level import → apply test with a native timeout
- pulling the 7-line refuse block pasted into all five ingests into a shared helper
- moving the hook helpers out of `keys.go` (which is otherwise about key quoting) into something like `adapter/hook_fields.go`

really close; the shape is right, and the codex fix alone is worth having. ping me once it's updated and I'll re-review 🚀

harryy2510:
@spxrogers thanks — that review was worth more than the PR was. Everything on the "To land" list is done, pushed as `8b442ad`, and both hermetic legs are green.

### The Gemini blocker

You were right, and it is worse than a wrong constant: the fixture I swapped at `gemini/apply_test.go` was ms-scale and I read it as seconds. The rule I broke is the one in CLAUDE.md about cross-referencing upstream docs before asserting what a harness does — I extrapolated Gemini's `timeout` from the other four adapters instead of checking. That it stayed green is the part worth fixing structurally, so:

- `adapter.ParseHookTimeoutMillis` / `SetHookTimeoutMillis` convert on both Gemini legs. The unit is in the **function name** rather than an argument, so every call site is right or wrong at a glance instead of depending on a constant three files away.
- A native value that is not a whole number of seconds is refused, not rounded. `1500` does not silently become 1s or 2s.
- `TestHookTimeout_CrossAdapterUnits` renders one canonical timeout through claude and gemini and requires the wire values to differ by exactly 1000x. That is the assertion a same-adapter round trip structurally cannot make, since both legs share the mistake and it cancels out. Plus on-disk assertions in both directions (`"timeout": 45000` out, Gemini's own `60000` default in as `60`).

### Structural vs semantic

Agreed that this regressed #124 from the other side. `ParseHookTimeout` now returns `source.HookTimeoutParse`: `OK`, `Unrepresentable` (a valid number the model cannot carry — semantic, still retires) and `Malformed` (not a number — structural, never retires). The third return value is gone; `Reason()` gives each refusal its own warning text, so "fix your typo" and "agentsync cannot model this" no longer read identically.

### The three silent drops

- **Explicit `0`** is refused. I took "refuse" over `*int` because `Timeout == 0` is already how the model spells "no timeout key", and a captured `0` re-renders with the key gone — swapping in the harness default. Written down on the `HookTimeoutUnrepresentable` const and in the capability matrix.
- **Negative canonical** fails the load naming the file, instead of being dropped by `SetHookTimeout`'s `> 0` guard.
- **Plugin projection** threads timeout through both manifest shapes, and `dedupHooks` keys on it. A plugin timeout the model cannot carry is logged and dropped rather than failing projection — deliberately unlike the canonical loader, since a third-party manifest must not be able to make itself unloadable. Reasoning is on `pluginHookTimeout`.

### One thing you flagged that was my error, not the code's

The PR description claimed `keys_test.go` covered `int64`, `float64` and fractional values. It did not — five subtests, none of them those. That is the "doc claims are not self-certifying" failure in the repo's own memory, committed in the document arguing the change was tested. The table is now real (every numeric shape × whole / fractional / negative / zero / overflow / NaN / non-numeric) and I rewrote the description to match what exists.

### Placement note worth your eye

The representability rule lives in `internal/source`, not `internal/adapter`, because `marketplace` needs it too and may not import `adapter` under the §12 layering graph. `adapter` keeps the native-map reading and the unit knowledge. If you would rather see the whole thing in `adapter/hook_fields.go` with a small duplicate in projection, say so and I will move it.

Hook helpers did move out of `keys.go` into `hook_fields.go` while I was in there, since `keys.go` is about key *names*.

### Not done, by your triage

- The CLI-level `import` → `apply` test with a native timeout.
- The 7-line refuse block is still pasted into all five ingests. `Reason()` / `Structural()` pulled the *wording* and the *classification* into one place, so the copies can no longer disagree about the interesting part, but the control flow is still five copies (each breaks a differently-labelled loop). Happy to do the real extraction if you want it in this PR.

Ready for another look.

